PR #30, opened 2024-11-12
Assign explicit message type values

apahl-cl:
In the `em_msg_type_t` enum from `em_base.h`, it looks like `em_msg_type_1905_ack` is out of order, causing the compiler to assign it a value of `0x8038` when the spec says 1905 Ack should be `0x8000`. For reference, here is a screenshot of the spec highlighted in red:

Screenshot 2024-11-12 at 12 14 21 PM

This doesn't present any symptoms/incorrect behavior because the codebase is internally consistent based on the values of this enum. If we test against devices running other EasyMesh implementations, we should expect to see conflicts when 1905 Ack is sent/received. 

This pull request fixes the 1905 Ack message value and explicitly assigns correct values to the rest of the enum members (in the hopes of avoiding future typos).

### DIFF
--- a/inc/em_base.h
+++ b/inc/em_base.h
@@ -301,7 +301,7 @@ typedef enum {
 } em_tlv_requirement_t;
 
 typedef enum {
-    em_msg_type_topo_disc,
+    em_msg_type_topo_disc = 0x0000,
     em_msg_type_topo_notif,
     em_msg_type_topo_query,
     em_msg_type_topo_resp,
@@ -312,7 +312,8 @@ typedef enum {
     em_msg_type_autoconf_resp,
     em_msg_type_autoconf_wsc,
     em_msg_type_autoconf_renew,
-    em_msg_type_ap_cap_query = 0x8001,
+    em_msg_type_1905_ack = 0x8000,
+    em_msg_type_ap_cap_query,
     em_msg_type_ap_cap_rprt,
     em_msg_type_map_policy_config_req,
     em_msg_type_channel_pref_query,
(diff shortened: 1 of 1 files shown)


amarnathhullur (approved):

amarnathhullur (changes requested):
Discussed internally on this change. It is better to have only the initial enum value of the block specified so that the natural property of enumeration to increment by 1 is not explicitly mentioned. Only if there is a jump in value the first base of that value is explicitly mentioned. 

Thus in the below case you could have explicitly value mentioned for em_msg_type_topo_disc=0x000 and em_msg_type_1905_ack=0x8000 and all other can sequentially get value without explicitly assignment. 

Could you please address this and update the Pull request.

apahl-cl:
Just pushed an update. There's also a jump at `em_msg_type_agent_list = 0x8035`, so note that I've explicitly set that value too.

amarnathhullur:
Hi, I see that in this pull request there are two commits, one assigning explicit value and other one removing that explicit value. Could you please raise a new pull request or update the pull request to only have one commit, which is addressing just the change of updating base value. It is little confusing to have two commits. 
Thanks.

apahl-cl:
Hey Amarnath, sure, I've squashed the commits. Let me know if I can do anything else.

amarnathhullur:
> Hey Amarnath, sure, I've squashed the commits. Let me know if I can do anything else.

Thanks, will check.

soumyasmunshi (approved):

PR #61, opened 2024-11-29
Fix for frequency band in EM

behanansaju:
Once the EM receives the radio frequency band from Onewifi, it will map the band to the corresponding frequency band and RF band based on the specifications.

### DIFF
--- a/inc/em_base.h
+++ b/inc/em_base.h
@@ -268,11 +268,18 @@ typedef enum {
 } em_profile_type_t;
 
 typedef enum {
-    em_freq_band_24,
+    em_freq_band_24,    //IEEE-1905-1-2013 table 6-23
     em_freq_band_5,
     em_freq_band_60,
     em_freq_band_unknown
 } em_freq_band_t;
+
+typedef enum {
+    em_rd_freq_band_unknown,    //rf band based on Wi-Fi Simple Configuration Technical Specification v2 table 44
+    em_rd_freq_band_24,
+    em_rd_freq_band_5,
+    em_rd_freq_band_60 = 0x4,
+} em_rd_freq_band_t;
 
 typedef struct {
     unsigned int    bit_map;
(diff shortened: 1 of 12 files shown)


inc/em_base.h
```diff
@@ -268,12 +268,19 @@ typedef enum {
 } em_profile_type_t;
 
 typedef enum {
-    em_freq_band_24,
+    em_freq_band_24,    //IEEE-1905-1-2013 table 6-23
     em_freq_band_5,
     em_freq_band_60,
```
amarnathhullur, line 273: what about 6G band? That needs to be considered as well.

soumyasmunshi (approved):

inc/em_configuration.h
```diff
@@ -102,8 +102,9 @@ class em_configuration_t {
     virtual void set_primary_device_type(char *) = 0;
     virtual em_freq_band_t get_band() = 0;
     virtual void set_band(em_freq_band_t band) = 0;
-
     virtual em_network_ssid_info_t *get_network_ssid_info_by_haul_type(em_haul_type_t haul_type) = 0;
+    virtual em_rd_freq_band_t map_freq_band_to_rf_band(em_freq_band_t band) = 0;
+    em_freq_band_t convert_freq_band(em_freq_band_t band);
```
amarnathhullur, line 107: input value and return value seems to be same enum. Is this correct?

src/em/em.cpp
```diff
@@ -517,6 +517,40 @@ dm_sta_t *em_t::find_sta(mac_address_t sta_mac, bssid_t bssid)
...
+        return em_rd_freq_band_24; // Return 2.4 GHz band
+    } else if (int_band == 1) {
+        return em_rd_freq_band_5; // Return 5 GHz band
+    } else if (int_band == 2) {
+        return em_rd_freq_band_60; // Return 60 GHz band
+    } else {
+        return em_rd_freq_band_unknown; // Return unknown for other values
+    }
+}
+
```
amarnathhullur, line 553: In both of the above functions, 6G should be taken into consideration.

behanansaju: 6 GHz seems to be not specified in the specification : Table 6-23—AutoconfigFreqBand TLV (EasyMesh v5 spec) , Table 44 – RF Bands( WSC specification v2)

inc/em_configuration.h
behanansaju: Yes.

PR #79, opened 2024-12-12
RDKBWIFI-13 Using std::vector for channels in em_op_class_info_t

marcin-matula-2329518:


inc/em_base.h
```diff
@@ -24,6 +24,8 @@
 #include <uuid/uuid.h>
 #include "ec_base.h"
 
+#include <vector>
```
marcin-matula-2329518, line 27: em_base.h is used in OneWifi which uses gcc to compile. vector cannot be used

marcin-matula-2329518: Hi @amarnathhullur yeap, I am looking for another solution for this segfault (please take notice that it is just draft for cli checking)

PR #90, opened 2024-12-17
RDKBWIFI-14 Prevent from setting value into index -1 (stack underflow) in dm_op_class_list.cpp

marcin-matula-2329518:


src/dm/dm_op_class_list.cpp
```diff
@@ -244,8 +244,9 @@ int dm_op_class_list_t::update_db(db_client_t& db_client, dm_orch_type_t op, voi
 		snprintf(tmp, sizeof(tmp), "%d,", info->channels[i]);
 		snprintf(channels_str + strlen(channels_str), sizeof(channels_str) - strlen(channels_str), "%s", tmp);
 	}
-
-	channels_str[strlen(channels_str) - 1] = 0;
+	if (strlen(channels_str) > 0) {
```
marcin-matula-2329518, line 247: To consider if better way is creating variable channels_str_strlen = strlen(channels_str) and using it into if conditon and assignment

soumyasmunshi (approved):

PR #102, opened 2024-12-22
CLI fixes

soumyasmunshi:


src/cli/em_cli.cpp
```diff
@@ -395,21 +415,132 @@ em_network_node_t *em_cli_t::get_network_tree(char *buff)
...
+	cloned->type = node->type;
+	strncpy(cloned->value_str, node->value_str, strlen(node->value_str) + 1);
+	cloned->value_int = node->value_int;
+
+	should_consider = (node->display_info.node_ctr == index);
+		
+	if (should_consider == true) {
+		if (collapse == false) {
+			tree_to_add = get_node_from_node_ctr(orig_node, node->display_info.orig_node_ctr);	
+			dump_lib_dbg(get_network_tree_string(tree_to_add));
```
amarnathhullur, line 502: Since get_network_tree_string is returing a malloced string, I think the return value should be freed after dump_lib_dbg call

PR #119, opened 2025-01-10
RDKBACCL-540: Observing segfault issue in em_ctrl when we tried to do reset command in em_cli prompt

kpandu987:
Reason for change: "haul_type_from_string() " api is not returning the type value .
Test Procedure: reset command is passed in cli
Risks: Low


manigandanj:
Looks okay to me.

amarnathhullur (approved):

PR #123, opened 2025-01-13
RDKBWIFI-13 Increase EM_MAX_CHANNELS_IN_LIST into 64.

marcin-matula-2329518:
Validation in db_easy_mesh_t::get_strings_by_token if number of value from string doesn't exceed limit (64).


src/db/db_easy_mesh.cpp
```diff
@@ -90,11 +90,12 @@ int db_easy_mesh_t::get_strings_by_token(char *parent, int token, unsigned int a
     while (tmp != NULL) {
         if ((tmp = strchr(orig, token)) != NULL) {
             *tmp = 0;
+            assert (num < argc - 1 && "number of extracted values exceeds the limit");
```
marcin-matula-2329518, line 93: Should be triggered this code and tested once more.

soumyasmunshi (approved):

PR #129, opened 2025-01-16
Added ability to "apply" certain Go CLI commands without updates + refactors

bcarlson-dev:
- Reset previously did not allow for "Apply" unless "Update" was used (a uniform requirement). If a network were in a fresh state, the "Update" would fail, not allowing the "Apply," which would, therefore, not allow for the intended "reset" behavior. Since this "unmodified apply" is ideal for at least one other command (provisioning), an additional struct parameter was added to change the skipping of "Apply." 
- Additional refactoring was done to reduce time complexity and improve readability without referencing the struct declaration.

**EDIT:** It appears VSCode applied Auto-Format to the main.go, which I didn't notice until now. I can revert if needed.

### DIFF
--- a/src/cli/main.go
+++ b/src/cli/main.go
@@ -28,207 +28,329 @@ import (
 
 const (
 	linesToDisplay int = 38
-	
-	NetworkTopologyCmd = "Network Topology"
-	NetworkSSIDListCmd = "SSID List"
-	RadioListCmd = "WiFi Radios"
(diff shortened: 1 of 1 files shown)


soumyasmunshi (approved):

bcarlson-dev:
@Nikita-Hakai

soumyasmunshi (approved):

PR #130, opened 2025-01-17
Fix for autoconfig renew looping & set_channel

behanansaju:


soumyasmunshi (changes requested):
Can you please delete the commented out code? There is no need of leaving the bad code.

PR #132, opened 2025-01-17
RDKBWIFI-10 Added support for btm_link_removal_imminent in em_steering_req_t.

marcin-matula-2329518:
It is related to delta between Wi-Fi_EasyMesh_Specification_v5.0.pdf and Wi-Fi_EasyMesh_Specification_v6.0.pdf specs:
_The Link Removal Imminent bit in the BTM Request frame shall be set to the value of the BTM_Link_Removal_Imminent bit in the Client Steering Request message received for the Steering Mandate_


marcin-matula-2329518:
@soumyasmunshi @Nikita-Hakai Hello, this change is only delta between 5.0 and 6.0 version of wifi mesh. It seems that merge those changes should not create any issue, but I leave you decision if and when it should be merged.

soumyasmunshi (approved):

PR #170, opened 2025-02-19
Initial refactoring and restructuring of EasyConnect code for DPP Onboarding

bcarlson-dev:


inc/ec_base.h
```diff
@@ -38,63 +38,123 @@ extern "C"
...
+    DPP_STATUS_RESPONSE_PENDING,
+    DPP_STATUS_INVALID_CONNECTOR,
+    DPP_STATUS_NO_MATCH,
+    DPP_STATUS_CONFIG_REJECTED,
+    DPP_STATUS_NO_AP_DISCOVERED,
+    DPP_STATUS_CONFIGURE_PENDING,
+    DPP_STATUS_CSR_NEEDED,
+    DPP_STATUS_CSR_BAD,
+    DPP_STATUS_NEW_KEY_NEEDED,
+} ec_status_code_t;
```
tuckerpo-cl, line 132: This enum is currently unused (is it necessary? for internal use, for is this a TLV member?), but a quick `to_str` for this with the `Meaning` column from the same table is probably useful for future logging / debugging.

inc/ec_session.h
```diff
@@ -22,49 +22,145 @@
...
+
+    /**
+     * @brief Create an authentication request frame in a pre-allocated buffer
+     * 
+     * @param buff The buffer to store the frame
+     * @return int The length of the frame
+     */
+    int create_auth_req(uint8_t *buff);
+
+    int handle_chirp_notification(em_dpp_chirp_value_t* chirp_tlv, uint8_t **out_frame);
```
tuckerpo-cl, line 128: Doc on this would be nice

inc/em_base.h
```diff
@@ -1702,14 +1700,10 @@ typedef enum {
...
-} em_hash_t;
-
-typedef struct {
-    unsigned char presence;
-    mac_address_t   enrollee[0];
-    em_hash_t   hash[0];
+    uint8_t mac_present : 1;  // Bit 7: Enrollee MAC Address Present
+    uint8_t hash_valid : 1;   // Bit 6: Hash Validity
+    uint8_t reserved : 6;     // Bits 5-0: Reserved
+    uint8_t data[];           // Flexible array for MAC address (if present) + hash length + hash value
```
tuckerpo-cl, line 1706: All other "flexible" arrays declare the final member explicitly as zero elements, i.e. `uint8_t data[0]` -- does this matter, or just stylistic?

src/em/crypto/em_crypto.cpp
```diff
@@ -699,6 +699,35 @@ uint8_t* em_crypto_t::base64_decode(const char* input, size_t length, size_t* ou
...
+    }
+
+    EC_KEY *ec_key = d2i_EC_PUBKEY(NULL, (const unsigned char **)&key, len);
+
+    EC_KEY_set_conv_form(ec_key, POINT_CONVERSION_COMPRESSED);
+    EC_KEY_set_asn1_flag(ec_key, OPENSSL_EC_NAMED_CURVE);
+
+    return ec_key;
+}
+
```
tuckerpo-cl, line 730: A few things on this:

- I'm fairly sure `EVP_DecodeBlock` dereferences its second argument, so add a `NULL` check on the input pubkey
- The `memset` call on `key` is redundant, since `EVP_DecodeBlock` will overwrite it anyways.
- After reading the openssl impl of `d2i_EC_PUBKEY`, it has a few codepaths that return `NULL`, so add a check before passing it to `EC_KEY_*`
- `(const unsigned char **)&key` makes me a bit uneasy, as I'm fairly sure it will decay to be `uint8_t(*)[1024]`  maybe usea separate pointer:
    `const unsigned char *p = &key; d2i_EC_PUBKEY(NULL, &p, len);`

src/em/prov/easyconnect/ec_session.cpp
```diff
@@ -0,0 +1,445 @@
...
+        m_params.responder_connector = EC_POINT_new(m_params.group);
+    }
+
+
+    m_params.x = BN_new();
+    m_params.y = BN_new();
+    m_params.m = BN_new();
+    m_params.n = BN_new();
+    m_params.prime = BN_new();
+    m_params.bnctx = BN_CTX_new();
```
tuckerpo-cl, line 209: I'm fairly sure all of these return `NULL` on failure, so add check / free prior to use

I think openSSL provides `BN_free` and `BN_CTX_free`, both of which handle being passed `NULL`, so no risk of something like the following:


```c
if (!m_params.x || !m_params.y || !m_params.m || !m_params.n || 
    !m_params.prime || !m_params.bnctx) {
    // error print
    BN_free(m_params.x);
    BN_free(m_params.y);
    BN_free(m_params.m);
    BN_free(m_params.n);
    BN_free(m_params.prime);
    BN_CTX_free(m_params.bnctx);
    return -1;
}
```

tuckerpo-cl (changes requested):
Overall looks good except for the few pieces of feedback. Additional non-critical nitpick: 2023 -> 2025 in copyright headers that you modified

inc/em_base.h
bcarlson-dev: So it's complicated. `[0]` is technically deprecated-ish and yields a warning (at least until one of the PRs I submitted is approved). Given the others are `[0]` I can switch to that.

src/em/crypto/em_crypto.cpp
bcarlson-dev: I'm updating, verify to see if this fits what you're seeing

inc/ec_base.h
bcarlson-dev: Good catch

tuckerpo-cl (approved):
LGTM 👍

PR #171, opened 2025-02-20
Removed "pedantic" flags when building

bcarlson-dev:
Removes the `-pedantic` and `-Wpedantic` flags from the build options.

These can be useful but given we are compiling a C++ project against two ANSI C projects, it will often result in many many warnings that **cannot** be resolved, either within those projects (`OneWifi`, `rdk-wifi-hal`) or within `em_base.h` which must include compatible C code for `OneWifi`. Having these just shows warnings that can't be fixed by `unified-wifi-mesh`.

```
/home/bcarlson/rdkb_easymesh/unified-wifi-mesh/build/ctrl/../../../OneWifi//include/wifi_base.h:290:24: warning: ISO C++ forbids zero-size array ‘frame_data’ [-Wpedantic]
  290 |     uint8_t frame_data[0];
```

### DIFF
--- a/build/agent/makefile
+++ b/build/agent/makefile
@@ -41,7 +41,7 @@ INCLUDEDIRS = \
 
 $(info The value of VARIABLE is $(ONEWIFI_HAL_INTF_HOME))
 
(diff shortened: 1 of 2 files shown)


bcarlson-dev:
@AswiniViswanathan @amarnathhullur 
I'm open to a discussion and arguments against this PR if I am missing anything

soumyasmunshi (approved):

PR #173, opened 2025-02-20
Removed "virtual" flags when building which conflict with **current** project architecture

bcarlson-dev:
I am open to discussion on this. Based on the current project structure where there is `em_t` which inherits from **virtual** classes `em_channel_t`, `em_capability_t`, `em_provisioning_t`, etc (which each have their own implementations), these warnings will unfortunately always be present. 

- If this architecture is open to change in the future, then I think these warnings can stay and this PR can be closed and **not merged**
- If this architecture will remain like this for the foreseeable future, then for clarity I think it would be best to disable them by accepting this PR.

Again this is open to discussion so please let me know how you feel:
@amarnathhullur @AswiniViswanathan

### DIFF
--- a/build/agent/makefile
+++ b/build/agent/makefile
@@ -41,7 +41,7 @@ INCLUDEDIRS = \
 
 $(info The value of VARIABLE is $(ONEWIFI_HAL_INTF_HOME))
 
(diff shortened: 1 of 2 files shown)


amarnathhullur:
@bcarlson-dev, I think disabling of these flags may hide potential issues. We are seeing runtime issues in Banana Pi with RDKB environment which are being addressed. Also, the next step is to address the various warnings seen and as part of that each warning should be checked if appropriate changes needs to be addressed. Thus, I would suggest to not enable these flags at this moment which will hide the warnings and potentially can cause issues in future.

bcarlson-dev:
@amarnathhullur 

Perfectly fine by me. This one was a bit iffy to me. I just wanted to put it out here.

PR #187, opened 2025-02-25
Implemented DPP Chirp Notification Handling and other EC enhancements

bcarlson-dev:
- Implemented the handling of DPP Chirp Notifications with the sending of the DPP Encap Messages in response.
- Added callable variables to `ec_session` to allow for exterior claling of send chirp and encap messages.
- Updated attribute methods to use heap allocation versus stack allocation.
- Replaced `set_auth_frame_wrapped_data` with a more universal `add_wrapped_data_attr`  which can be used whenever authentication is needed by having the caller provide their own key and attributes to wrap.
- Added static `em_msg_t` utilities to reduce the code duplication in Controller and Agent code (still not applied everywhere)

### DIFF
--- a/src/em/prov/easyconnect/ec_session.cpp
+++ b/src/em/prov/easyconnect/ec_session.cpp
@@ -371,6 +371,10 @@ int ec_session_t::handle_chirp_notification(em_dpp_chirp_value_t *chirp_tlv, uin
 
 }
 
+int ec_session_t::handle_proxy_encap_dpp_tlv(em_encap_dpp_t *encap_tlv, uint8_t **out_frame) {
+
+}
+
 int ec_session_t::set_auth_frame_wrapped_data(ec_frame_t *frame, unsigned int non_wrapped_len, bool do_init_auth)
 {
     siv_ctx ctx;
@@ -448,7 +452,9 @@ int ec_session_t::handle_recv_ec_action_frame(ec_frame_t *frame, size_t len)
     return 0;
 }
 
-ec_session_t::ec_session_t()
+ec_session_t::ec_session_t(std::function<int(em_dpp_chirp_value_t*, size_t)> send_chirp_notification,
+                            std::function<int(em_encap_dpp_t*, size_t, em_dpp_chirp_value_t*, size_t)> send_prox_encap_dpp_msg)
+                            : m_send_chirp_notification(send_chirp_notification), m_send_prox_encap_dpp_msg(send_prox_encap_dpp_msg)
 {
     // Initialize member variables
     m_cfgrtr_ver = 0;
--- a/src/em/prov/easyconnect/ec_session.cpp
+++ b/src/em/prov/easyconnect/ec_session.cpp
@@ -24,89 +24,103 @@
 #include "em.h"
 #include "aes_siv.h"
 
-int ec_session_t::create_auth_req(uint8_t *buff)
+std::pair<uint8_t*, uint16_t> ec_session_t::create_auth_request()
 {
 
     EC_KEY *responder_boot_key, *initiator_boot_key;
-    unsigned int wrapped_len;
 
-    uint16_t attrib_len, chann_attr;;
-    uint8_t protocol_key_buff[1024];
+    ec_dpp_capabilities_t caps = {{
+        .enrollee = 0,
+        .configurator = 1
+    }};
 
     printf("%s:%d Enter\n", __func__, __LINE__);
 
+    uint8_t* buff = (uint8_t*) calloc(EC_FRAME_BASE_SIZE, 1);
+
     ec_frame_t    *frame = (ec_frame_t *)buff;
-
-    attrib_len = 0;
-
     frame->frame_type = ec_frame_type_auth_req;
 
     if (init_session(NULL) != 0) {
         m_activation_status = ActStatus_Failed;
         printf("%s:%d Failed to initialize session parameters\n", __func__, __LINE__);
-        return -1;
+        return std::make_pair<uint8_t*, uint16_t>(NULL, 0);
     }
 
     if (compute_intermediate_key(true) != 0) {
         m_activation_status = ActStatus_Failed;
         printf("%s:%d failed to generate key\n", __func__, __LINE__);
-        return -1;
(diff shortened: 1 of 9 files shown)


src/em/prov/easyconnect/ec_session.cpp
```diff
@@ -351,39 +366,71 @@ int ec_session_t::handle_chirp_notification(em_dpp_chirp_value_t *chirp_tlv, uin
     memcpy(hash, data_ptr, hash_len);
 
     // Validate hash
-    if (compute_key_hash(m_data.responder_boot_key, m_params.responder_keyhash) < 1) {
+    // Compute the hash of the responder boot key 
+    uint8_t resp_boot_key_chirp_hash[SHA512_DIGEST_LENGTH];
```
tuckerpo-cl, line 370: Should be `SHA256_DIGEST_LENGTH`, right? Nitpicking because it's just buffer length (`compute_key_hash` still correctly uses SHA256)

src/em/prov/easyconnect/ec_session.cpp
```diff
@@ -351,39 +366,71 @@ int ec_session_t::handle_chirp_notification(em_dpp_chirp_value_t *chirp_tlv, uin
...
-    // TODO: 
-    // create_auth_req(*out_frame);
+    auto [auth_frame, auth_frame_len] = create_auth_request();
+    if (auth_frame == NULL || auth_frame_len == 0) {
+        printf("%s:%d: Failed to create authentication request frame\n", __func__, __LINE__);
+        return -1;
+    }
+
+    // Create Auth Request Encap TLV: EasyMesh 5.3.4
+    em_encap_dpp_t* encap_dpp_tlv = (em_encap_dpp_t*)calloc(sizeof(em_encap_dpp_t) + auth_frame_len , 1);
```
tuckerpo-cl, line 391: I'd check that `calloc` doesn't fail here

src/em/prov/easyconnect/ec_session.cpp
```diff
@@ -351,39 +366,71 @@ int ec_session_t::handle_chirp_notification(em_dpp_chirp_value_t *chirp_tlv, uin
...
+
+    memcpy(encap_dpp_tlv->dest_mac_addr, mac, sizeof(mac_addr_t));
+    encap_dpp_tlv->encap_frame_len = auth_frame_len;
+    memcpy(encap_dpp_tlv->encap_frame, auth_frame, auth_frame_len);
+
+    free(auth_frame);
+
+    // Create Auth Request Chirp TLV: EasyMesh 5.3.4
+    size_t data_size = sizeof(mac_addr_t) + hash_len + sizeof(uint8_t);
+    em_dpp_chirp_value_t* chirp = (em_dpp_chirp_value_t*)calloc(sizeof(em_dpp_chirp_value_t) + data_size, 1);
```
tuckerpo-cl, line 404: `calloc` failure check as well

src/em/prov/em_provisioning.cpp
```diff
@@ -100,79 +101,114 @@ int em_provisioning_t::create_cce_ind_msg(unsigned char *buff)
...
-    tmp += sizeof(mac_address_t);
-    len += sizeof(mac_address_t);
+    //dm_easy_mesh_t *dm = get_data_model();
 
-    memcpy(tmp, (unsigned char *)&type, sizeof(unsigned short));
-    tmp += sizeof(unsigned short);
-    len += sizeof(unsigned short);
+    //TODO: Decide on addressing.
+    //tmp = em_msg_t::add_1905_header(tmp, &len, dm->get_agent_al_interface_mac(), dm->get_ctrl_al_interface_mac(), em_msg_type_proxied_encap_dpp);
+    tmp = em_msg_t::add_1905_header(tmp, &len, (uint8_t *)get_peer_mac(), get_current_cmd()->get_al_interface_mac(), em_msg_type_proxied_encap_dpp);
```
tuckerpo-cl, line 119: Looking for clarification: is this written from the perspective of the Proxy Agent? I'm still a smidge confused as to how we're going to (at runtime) determine who "we" are and who to forward things to. Because if this is running in an Enrollee, the dest addr is going to be the PA, but if running as PA, dest addr would be controller :thinking:

tuckerpo-cl (changes requested):

src/em/prov/easyconnect/ec_session.cpp
bcarlson-dev: That was from the existing implementation, with a lot of hard coded hashes placed (even though it should allow for a "user specified" prime -> hash). I am planning to address all of those in later PR where the hash functions and more are correctly aligned with the passed function and the spec. I haven't quite got there yet but I know I'll run into it soon.

src/em/prov/em_provisioning.cpp
bcarlson-dev: So no matter what (as alluded to by the TODO), addressing is something I am looking more into since there is the question of "which agent" when going from controller to agent that will be a bit more complex. Not to mention the Rust library _should_ just make this work soon so there's a bit of uncertainty. 

Regarding the Proxy Agent vs Enrollee, only the PA can send 1905 messages to the controller while DPP is happening. Once the Enrollee is onboarded to the network then it can send 1905 messages but at that point it's no longer an enrollee. So if it's an agent (which you can see by `get_service_type()`) we have to send it to the controller, since that agent could only be the PA. If it's the controller though...I'll see what that's like when I get there

src/em/prov/easyconnect/ec_session.cpp
bcarlson-dev: Good catch, I try but I keep missing some of them

src/em/prov/em_provisioning.cpp
tuckerpo-cl: Yeah, we should probably begin incorporating `switch (get_service_type())` calls before statefulness of enrollee / PA / Controller blows up in complexity.
BTW, I think `get_service_type` probably needs a `proxy_agent` enum, or maybe just agent and a `is_provisioned` flag where `is_provisioned == false` indicates that we're a PA is good enough.

tuckerpo-cl (approved):

PR #189, opened 2025-02-26
build/cli/makefile: fix run target

tuckerpo-cl:
Fixes the `run` target in the CLI `makefile`

Fixes both path to target binary `em_onewifi_cli` and also includes `LD_LIBRARY_PATH` as the dependent `libemcli.so` library is placed in `install/lib/` as opposed to `/usr/*/lib/` or the system `$PATH`

Also adds `RUNARGS` to be able to pass arguments to the CLI binary.

Example invocation if running on a pi:
```sh
make run RUNARGS=rpi
```

### DIFF
--- a/build/cli/makefile
+++ b/build/cli/makefile
@@ -105,10 +105,10 @@ clean:
 
 #
 # Run target: "make -f Makefile.Linux run" to execute the application
-#             You will need to add $(VARIABLE_NAME) for any command line parameters 
+#             You will need to add $(RUNARGS) for any command line parameters
 #             that you defined earlier in this file.
 # 
 
 run:
-	./$(PROGRAM) 
+	LD_LIBRARY_PATH=$(INSTALLDIR)/lib $(PROGRAM) $(RUNARGS)
 


bcarlson-dev:
Can you do a verification and change directory to the install directory before running? There are certain code segments that assume you are running from that install directory (given the relative file paths) so it might prevent some issues with commands not working when it's called from elsewhere in the project structure

tuckerpo-cl:
@bcarlson-dev The makefiles already have a hard-coded file path dependency in terms of `makefile.inc` found within `build/`, example:

```sh
tpolomik@raspberrypi:~/rdk-easymesh-dev/unified-wifi-mesh $ make -f build/cli/makefile run RUNARGS=rpi
build/cli/makefile:19: ../makefile.inc: No such file or directory
make: *** No rule to make target '../makefile.inc'.  Stop.
```
Or, just the `all` target:
```sh
tpolomik@raspberrypi:~/rdk-easymesh-dev/unified-wifi-mesh $ make -f build/cli/makefile
build/cli/makefile:19: ../makefile.inc: No such file or directory
make: *** No rule to make target '../makefile.inc'.  Stop.
```
So the `makefile`(s) are kinda busted unless you're running from `build/<folder>/` anyways.

bcarlson-dev:
Makes sense then LGTM

bcarlson-dev (approved):

PR #196, opened 2025-03-01
Draft: WIP: GAS frame handling for DPP

tuckerpo-cl:
@bcarlson-dev I'd like to get your take on this before going further. What's your take on my current impl of forwarding GAS frames to `ec_session` to handle? I imagine we could add another `std::function<(...)>`, like `send_gas_response`, that's bound to an implementation in `em_provisioning`, but a bit unsure on how to handle states and such.

For example, in `ec_session::handle_gas_initial_req`, we'll need to know if we're the Configurator or Enrollee, if Enrollee just ignore, and if Configurator, read relevant information and craft a reply and send it back to Enrollee via aforementioned hypothetical `send_gas_response` function pointer, but TMK that's all encapsulated within `em_provisioning` itself. Also unsure on the best way to proceed for _sending_ GAS frames, from the Enrollee's PoV. Thoughts?

### DIFF
--- a/inc/ec_base.h
+++ b/inc/ec_base.h
@@ -42,6 +42,7 @@ extern "C"
 
 #define DPP_OUI_TYPE 0x1A
 #define DPP_MAX_EN_CHANNELS 4
+static const uint8_t WFA_DPP_OUI[3] = {0x50, 0x6F, 0x9A};
 
 // As defined by EasyConnect 8.2.1 Table 35
 typedef enum  {
--- a/inc/ec_session.h
+++ b/inc/ec_session.h
@@ -28,6 +28,9 @@
 #include <functional>
 #include <vector>
 
+// forward decl
+struct ieee80211_mgmt;
+
 #define EC_FRAME_BASE_SIZE (offsetof(ec_frame_t, attributes))
 
 class ec_session_t {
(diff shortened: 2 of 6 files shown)


bcarlson-dev:
That's a very interesting question about the sending of 802.11 frames (both DPP Action and GAS frames)  from the `ec_session`. I think you are right that it would be best to have another callable `std::function` for those but that also brings up some additional comments/questions. 
1. `m_bus_hdl` (`em_agent.cpp`) is the only registered bus handle which I think should stay that way to avoid confusion. It also makes it clear that there is only communication with the bus from the agent's POV (since that's where OneWifi is running from).
2. Given that, we need to define the `send_action_frame` and `send_gas_frame` functions in `em_agent.cpp`, make them call `Device.WiFi.AccessPoint.1.RawFrame.Mgmt.Action.Tx` and then give those std::functions to the `ec_session` in all of the `em_t`s on that agent.

On a side note, I think it's also important that the `send_action_frame` and `send_gas_frame` functions be located in `em_agent` because these will likely be used outside of EasyConnect in someway so having that more accessible and just passing it through would be a plus.

Regarding the deciding on the state between Initiator/Configurator and Responder/Enrollee, no matter what I think we have to create a new enum that is contained entirely within the `ec_session`. Currently all the `em_provisioning/em_t` knows is if it's an agent or a controller but the agent itself can be an enrollee or a configurator so we need to add more.

tuckerpo-cl:
Closing in favor of #207

PR #199, opened 2025-03-03
RDKBACCL-618 : obsering incorrect mac address

kpandu987:
Reason for change:  Adding bpi support to fetch proper erouter mac address for ctrl
Test Procedure: Able to see proper mac addr in ctrl 
"init_network_topology:1879: Root: 0e:e5:d7:73:c5:e5  added to network topology"
 Risks: Low

### DIFF
--- a/install/bin/Reset.json
+++ b/install/bin/Reset.json
@@ -10,7 +10,9 @@
 					"rpi": "eth"
 				}, {
 					"sim": "ens"
-			}],
+				}, {
+					"bpi": "erouter"
+				}],
 			"List": []
 		},
         "NetworkSSIDList": [{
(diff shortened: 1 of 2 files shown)


install/bin/Reset.json
```diff
@@ -10,6 +10,8 @@
 					"rpi": "eth"
 				}, {
 					"sim": "ens"
+			}, {
```
manigandanj, line 13: Please check the indent.

kpandu987: Taken care

install/bin/Reset.json
```diff
@@ -10,7 +10,9 @@
 					"rpi": "eth"
 				}, {
 					"sim": "ens"
-			}],
+				}, {
+					"bpi": "erouter"
```
amarnathhullur, line 14: why do we need to take erouter for bpi, can it not be eth itself as that of rpi?

cpokuru: In BPI lan0 will be renamed to erouter0

soumyasmunshi (approved):

PR #202, opened 2025-03-05
Fix for warnings treated as error in the ctrl - part 4

behanansaju:


src/em/config/em_configuration.cpp
```diff
@@ -3334,11 +3314,12 @@ em_wsc_msg_type_t em_configuration_t::get_wsc_msg_type(unsigned char *buff, unsi
 
 int em_configuration_t::handle_ap_radio_advanced_cap(unsigned char *buff, unsigned int len)
 {
-    dm_easy_mesh_t *dm;
-    dm = get_data_model();
+    //dm_easy_mesh_t *dm;
+    //dm = get_data_model();
```
amarnathhullur, line 3318: remove the commented code, if not used

amarnathhullur (approved):

PR #204, opened 2025-03-05
EasyConnect Seperation of Concerns Refactor

bcarlson-dev:
- Separated Controller, Proxy Agent, and Enrollee into different classes for better code understanding
- Introduced `ec_manager`


inc/ec_configurator.h
```diff
@@ -0,0 +1,179 @@
...
+
+    std::string m_mac_addr;
+
+    // The connections to the Enrollees/Agents
+    std::map<std::string, ec_connection_context_t> m_connections;
+
+    inline ec_connection_context_t& get_conn_ctx(const std::string& mac) {
+        // TODO: Add a check for the connection
+        return m_connections[mac];    
+    }
```
tuckerpo-cl, line 168:
```c++
auto it = m_connections.find(mac);
if (it != m_connections.end()) return it->second;
return ???
```
Maybe change to `ec_connection_context_t *` return so the case where the MAC is unknown can be handled

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -0,0 +1,369 @@
...
+    // Validate hash
+    // Compute the hash of the responder boot key 
+    uint8_t *resp_boot_key_chirp_hash = ec_crypto::compute_key_hash(m_boot_data.responder_boot_key, "chirp");
+    if (resp_boot_key_chirp_hash == NULL) {
+        printf("%s:%d unable to compute \"chirp\" responder bootstrapping key hash\n", __func__, __LINE__);
+        return false;
+    }
+
+    if (memcmp(hash, resp_boot_key_chirp_hash, hash_len) != 0) {
+        // Hashes don't match, don't initiate DPP authentication
```
tuckerpo-cl, line 28: `free(resp_boot_key_chirp_hash);`

tuckerpo-cl (changes requested):
Stylistically I'd like to see `override` used in subclasses where applicable, just makes reasoning about the code easier for instance if you're reading a subclass header, immediately obvious that a given function extends some other base class. Also, regarding `send_action_frame` as a param for `ec_manager`, would that be bound to `send_frame`, or some new entry into `em_agent` which sends action frames via the bus to `onewifi`? Other than this comment and the other 2 left, LGTM.

inc/ec_configurator.h
bcarlson-dev: Sounds good. I was hoping to avoid adding "yet another NULL check" but sometimes you can't help it

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
bcarlson-dev: Good catch

bcarlson-dev:
> Stylistically I'd like to see `override` used in subclasses where applicable, just makes reasoning about the code easier for instance if you're reading a subclass header, immediately obvious that a given function extends some other base class. Also, regarding `send_action_frame` as a param for `ec_manager`, would that be bound to `send_frame`, or some new entry into `em_agent` which sends action frames via the bus to `onewifi`? Other than this comment and the other 2 left, LGTM.

I added the `override` and totally agree. I meant to do it but got mixed up in the fact that C++ doesn't _require_ it. 

Regarding the `send_action_frame` that'll need to be a new function definition in `em_agent` that utilizes the OneWifi `"Device.WiFi.AccessPoint.{i}.RawFrame.Mgmt.Action.Tx"` call.

tuckerpo-cl (approved):

PR #207, opened 2025-03-06
Add basic DPP config / GAS frame message infra

tuckerpo-cl:
Forward DPP Config Req/Resp (GAS frames) to 'ec_manager'

Stub dummy handlers in ec_pa_configurator and ec_enrollee

### DIFF
--- a/inc/em_base.h
+++ b/inc/em_base.h
@@ -209,6 +209,10 @@ extern "C"
 #define 	EM_PARSE_ERR_CONFIG		EM_PARSE_NO_ERR	- 3	
 #define 	EM_PARSE_ERR_NO_CHANGE	EM_PARSE_NO_ERR	- 4	
 
+#ifndef ETH_ALEN
+#define ETH_ALEN (6UL)
+#endif // ETH_ALEN
+
 typedef char em_interface_name_t[32];
 typedef unsigned char em_nonce_t[16];
 typedef unsigned char em_dh5_key_t[192];    // because this is DH group 5 (1536 bits)
@@ -2576,7 +2580,8 @@ typedef enum {
     em_bus_event_type_get_mld_config,
     em_bus_event_type_mld_reconfig,
     em_bus_event_type_beacon_report,
-    em_bus_event_type_recv_wfa_action_frame
+    em_bus_event_type_recv_wfa_action_frame,
+    em_bus_event_type_recv_gas_frame,
 } em_bus_event_type_t;
 
 typedef struct {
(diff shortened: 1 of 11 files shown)


src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -166,7 +166,26 @@ bool ec_enrollee_t::handle_auth_confirm(uint8_t *buff, unsigned int len)
 
 bool ec_enrollee_t::handle_config_response(uint8_t *buff, unsigned int len)
 {
+#define APEFMT "%02x,%02x,%02x"
+#define APE2STR(x) (x[0], x[1], x[2])
+#define APEIDFMT "%02x,%02x,%02x,%02x,%02x,%02x,%02x"
+#define APEID2STR(x) (x[0], x[1], x[2], x[3], x[4], x[5], x[6])
+    uint8_t *p = buff;
```
bcarlson-dev, line 173: I'm guessing you didn't want to define these in a header because of the global scope and their sort of limited use case/naming, would putting them in `ec_util.h` help with that? It _"should"_ only be used inside `ec_*` files and it's not included globally like the `ec_base.h` is.

bcarlson-dev (approved):
Beyond that option it LGTM

src/em/prov/easyconnect/ec_enrollee.cpp
tuckerpo-cl: Honestly it was just scoped inside this function for stub purposes, but I'll move it into `ec_base.h` as a helper for debugging

inc/em_base.h
```diff
@@ -209,6 +209,10 @@ extern "C"
 #define 	EM_PARSE_ERR_CONFIG		EM_PARSE_NO_ERR	- 3	
 #define 	EM_PARSE_ERR_NO_CHANGE	EM_PARSE_NO_ERR	- 4	
 
+#ifndef ETH_ALEN
```
bcarlson-dev, line 212: I found that `ETH_ALEN` is defined in [ieee80211.h](https://github.com/rdkcentral/unified-wifi-mesh/blob/main/inc/ieee80211.h#L1463) if you want to include that instead

bcarlson-dev (approved):
Besides that it LGTM

inc/em_base.h
bcarlson-dev: (not included in `em_base.h` since that could create some conflicts with OneWifi but included where it's needed)

PR #213, opened 2025-03-11
DPP EasyConnect: Fixed all non-deprecated warnings

bcarlson-dev:
- Added C++ style casts to appropriate places
- Used `new` and `delete` when allocated `uint8_t` arrays to reduce casts
- Reformatted some args to reduce casts
- Changed some size typing to reflect spec requirements:
    * attribute max size can be `uint16_t`
    * many attributes can be added into a `size_t` buffer (could be fragmented)

### DIFF
--- a/src/em/prov/easyconnect/ec_ctrl_configurator.cpp
+++ b/src/em/prov/easyconnect/ec_ctrl_configurator.cpp
@@ -11,7 +11,7 @@ bool ec_ctrl_configurator_t::process_chirp_notification(em_dpp_chirp_value_t *ch
     uint8_t hash[255] = {0}; // Max hash length to avoid dynamic allocation
     uint8_t hash_len = 0;
 
-    if (ec_util::parse_dpp_chirp_tlv(chirp_tlv, tlv_len, &mac, reinterpret_cast<uint8_t**> (&hash), &hash_len) == false) {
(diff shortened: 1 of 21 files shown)


src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -606,71 +602,70 @@ std::pair<uint8_t *, uint16_t> ec_ctrl_configurator_t::create_auth_confirm(std::
...
     uint8_t* attribs = NULL;
-    uint16_t attrib_len = 0;
+    size_t attribs_len = 0;
 
     // TODO: Add transaction ID outside this function
     uint8_t trans_id = 0;
     ec_dpp_reconfig_flags_t reconfig_flags = {
         .connector_key = 1, // DONT REUSE
-        .reserved = 7
+        .reserved = 0
```
amarnathhullur, line 637: is this ok to change from 7 to 0. Why was this 7 earlier?

amarnathhullur (approved):
All other changes looks good.

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
bcarlson-dev: Sorry I didn't see this comment before I merged. That was a mistake that was just being fixed up. Theoretically reserved shouldn't matter but given people can implement the spec however they want it's probably best to just set it to 0.

PR #216, opened 2025-03-12
EC 6.4.2 DPP Configuration Object creation + send from Enrollee, GAS frame helpers

tuckerpo-cl:
Enrollee only ATM

@bcarlson-dev This needs to be rebased onto `main` but putting up as draft to hopefully discuss what we ought to do for Configuration Request Object creation (see: `4.4 DPP Configuration Request Object` in EasyConnect and `Table 5. DPP Configuration Request object` in EasyMesh), since we need fairly in-depth radio configuration information which `ec_enrollee.h` obviously doesn't have. I was thinking either a bound std::function to _some_ `em_t` which can just return a cJSON object potentially, or maybe a callback bound to some OneWifi call that gets radio info? Unsure what the "right" approach would be and I'd appreciate your thoughts

### DIFF
--- a/inc/ec_util.h
+++ b/inc/ec_util.h
@@ -133,6 +133,48 @@ public:
         return frame;
     }
 
+    static std::pair<void *, size_t> alloc_gas_frame(dpp_gas_action_type_t action, uint8_t dialog_token) {
+        void *frame = nullptr;
+        size_t created_frame_size = 0UL;
+        switch(action) {
+            case dpp_gas_action_type_t::dpp_gas_initial_req: {
+                frame = calloc(1, sizeof(ec_gas_initial_request_frame_t));
+                if (!frame) {
+                    printf("%s:%d: Failed to allocate GAS frame!\n", __func__, __LINE__);
+                    break;
+                }
+                auto *req_frame = static_cast<ec_gas_initial_request_frame_t *>(frame);
+                memcpy(req_frame->ape, DPP_GAS_CONFIG_REQ_APE, sizeof(req_frame->ape));
+                memcpy(req_frame->ape_id, DPP_GAS_CONFIG_REQ_PROTO_ID, sizeof(req_frame->ape_id));
+                created_frame_size = sizeof(ec_gas_initial_request_frame_t);
+            }
+            break;
+            case dpp_gas_action_type_t::dpp_gas_initial_resp: {
+                frame = calloc(1, sizeof(ec_gas_initial_response_frame_t));
+                if (!frame) {
+                    printf("%s:%d: Failed to allocate GAS frame!\n", __func__, __LINE__);
+                    break;
+                }
+                auto *resp_frame = static_cast<ec_gas_initial_response_frame_t *>(frame);
+                memcpy(resp_frame->ape, DPP_GAS_CONFIG_REQ_APE, sizeof(resp_frame->ape));
+                memcpy(resp_frame->ape_id, DPP_GAS_CONFIG_REQ_PROTO_ID, sizeof(resp_frame->ape_id));
+                created_frame_size = sizeof(ec_gas_initial_response_frame_t);
+            }
+            break;
+            default:
+                printf("%s:%d: unhandled GAS frame type=%02x\n", __func__, __LINE__, action);
+                break;
+        }
+        // Shared fields
+        if (frame) {
+            ec_gas_frame_base_t *base = static_cast<ec_gas_frame_base_t *>(frame);
+            base->category = 0x04;
+            base->action = static_cast<uint8_t>(action);
+            base->dialog_token = dialog_token;
+        }
+        return std::make_pair(frame, created_frame_size);
+    }
+
     /**
      * @brief Copy (overrride) attributes to a frame
      * 
@@ -144,6 +186,17 @@ public:
      * @warning The frame must be freed by the caller
      */
     static ec_frame_t* copy_attrs_to_frame(ec_frame_t *frame, uint8_t *attrs, size_t attrs_len);
+
+    /**
+     * @brief Copy (over-write) attributes to a frame
+     * 
+     * @param frame The frame to copy the attribues to
+     * @param frame_base_size The offset at which to copy attributes to
+     * @param attrs The attributes to copy
+     * @param attrs_len The length of the attributes
+     * @return uint8_t *, base of the frame with newly copied attributes, or nullptr on failure
+     */
+    static uint8_t* copy_attrs_to_frame(uint8_t *frame, size_t frame_base_size, uint8_t *attrs, size_t attrs_len);
 
     /**
      * @brief Validate an EC frame based on the WFA parameters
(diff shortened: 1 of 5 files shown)


bcarlson-dev:
So, looking at **EasyConnect**, the main outside information dependency is SSID/password/auth type information (in Object+Responses). In **EasyMesh** that's where and everything else comes into play.

I think that adding additional `std::function` variables is a good idea and that it should return a `cJSON` object or an `std::string` (whichever). That way (especially since EasyMesh is changing and will continue to change when it comes to radio dynamics and MLO), it's easy to modify the json in that method versus having to continually add more and more callbacks. I think it could also take the `ec_connection_context_t` as a parameter in order to have access to the connectors and keys that `ec_*` are generating.

To summarize: 
I think callables like this: 
```cpp
using generate_em_config_req = std::function<cJSON*(ec_connection_context_t&)>;
using generate_em_config_obj = std::function<cJSON*(ec_connection_context_t&)>;
// OR
`using generate_em_config_req = std::function<std::string(ec_connection_context_t&)>`;
`using generate_em_config_obj = std::function<std::string(ec_connection_context_t&)>`;
```
would be good, with the actual implementations being either located in the `em_agent_t` or the `em_ctrl_t` (depending on context).

The implementations could then use the `m_data_model` member in those classes to get the data needed, which, along with the `ec_connection_context_t` data would create the JSON object for the `ec_*` classes to use.

bcarlson-dev:
I still can't nail down whether EasyMesh just adds on another Configuration Request Object and another Configuration Object or if it replaces the EasyConnect one. 

If it replaces it, then those functions could just be `generate_config_req_fn` and `generate_config_obj` (or something similar). If it is in addition to it, then I think we need four of them total

tuckerpo-cl:
> I still can't nail down whether EasyMesh just adds on another Configuration Request Object and another Configuration Object or if it replaces the EasyConnect one.
> 
> If it replaces it, then those functions could just be `generate_config_req_fn` and `generate_config_obj` (or something similar). If it is in addition to it, then I think we need four of them total

I **believe** that EasyMesh is in addition to the EasyConnect definition, with one request JSON blob per _thing_ being configured, but I'm also not 100% certain about that.

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -301,6 +303,19 @@ bool ec_enrollee_t::handle_auth_confirm(ec_frame_t *frame, size_t len, uint8_t s
         return false;
     }
```
bcarlson-dev, line 305: Can you just add a bit of in-depth documentation here saying that according to the spec *this* is the point when we are switching from the authentication stage to the configuration stage?

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -571,7 +586,81 @@ std::pair<uint8_t *, size_t> ec_enrollee_t::create_recfg_auth_response(ec_status
...
+    attribs = ec_util::add_wrapped_data_attr((uint8_t *)initial_req_frame, sizeof(ec_gas_initial_request_frame_t), attribs, &attribs_len, true, m_eph_ctx.ke, [&](){
+        size_t wrapped_len = 0;
+        uint8_t* wrapped_attribs = ec_util::add_attrib(nullptr, &wrapped_len, ec_attrib_id_enrollee_nonce, m_p_ctx.nonce_len, m_eph_ctx.e_nonce);
+        wrapped_attribs = ec_util::add_attrib(wrapped_attribs, &wrapped_len, ec_attrib_id_dpp_config_req_obj, config_obj_len, (uint8_t *)dpp_config_request_obj);
+        return std::make_pair(wrapped_attribs, wrapped_len);
+    });
+
+    if ((initial_req_frame = reinterpret_cast<ec_gas_initial_request_frame_t*>(
+        ec_util::copy_attrs_to_frame((uint8_t*)initial_req_frame, sizeof(ec_gas_initial_request_frame_t), attribs, attribs_len))) == nullptr) {
+        printf("%s:%d: unable to copy attribs to GAS frame\n", __func__, __LINE__);
```
bcarlson-dev, line 658: `attribs`, and `frame` need to be `free`d here

inc/ec_util.h
```diff
@@ -133,6 +133,48 @@ class ec_util {
...
+        void *frame = nullptr;
+        size_t created_frame_size = 0UL;
+        switch(action) {
+            case dpp_gas_action_type_t::dpp_gas_initial_req: {
+                frame = calloc(1, sizeof(ec_gas_initial_request_frame_t));
+                if (!frame) {
+                    printf("%s:%d: Failed to allocate GAS frame!\n", __func__, __LINE__);
+                    break;
+                }
+                auto *req_frame = (ec_gas_initial_request_frame_t *)frame;
```
bcarlson-dev, line 146: Does this need a `reinterpret_cast` or does the `auto` get around that?

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -301,6 +303,19 @@ bool ec_enrollee_t::handle_auth_confirm(ec_frame_t *frame, size_t len, uint8_t s
...
+    }
+
+    if (m_send_action_frame(src_mac, config_req, config_req_len, 0)) {
+        printf("%s:%d: Sent DPP Configuration Request 802.11 frame to Proxy Agent!\n", __func__, __LINE__);
+    } else {
+        printf("%s:%d: Failed to send DPP Configuration Request GAS frame\n", __func__, __LINE__);
+        return false;
+    }
+
     return true;
```
bcarlson-dev, line 319: Make sure to free `config_req` in the `return false` and `return true` case

src/em/prov/easyconnect/ec_util.cpp
```diff
@@ -95,6 +95,63 @@ bool ec_util::validate_frame(const ec_frame_t *frame)
...
+    uint8_t* ret_frame_attribs = ec_util::add_attrib(frame_attribs, non_wrapped_len, ec_attrib_id_wrapped_data, wrapped_attrib_len, reinterpret_cast<uint8_t *>(wrapped_attrib));
+
+
+    free(wrap_attribs);
+
+    return ret_frame_attribs;
+
+}
 
 uint8_t* ec_util::add_wrapped_data_attr(ec_frame_t *frame, uint8_t* frame_attribs, size_t* non_wrapped_len, bool use_aad, uint8_t* key, std::function<std::pair<uint8_t*, uint16_t>()> create_wrap_attribs)
```
bcarlson-dev, line 156: Can you make it so this function just calls the new `uint8_t* add_wrapped_data_attr` function (although with the casts and sizeof)

src/em/prov/easyconnect/ec_util.cpp
```diff
@@ -331,6 +388,17 @@ ec_frame_t *ec_util::copy_attrs_to_frame(ec_frame_t *frame, uint8_t *attrs, size
 
 }
 
+uint8_t *ec_util::copy_attrs_to_frame(uint8_t *frame, size_t frame_base_size, uint8_t *attrs, size_t attrs_len) {
```
bcarlson-dev, line 391: Likewise, can you make it so the `ec_frame_t* copy_attrs_to_frame` calls the `uint8_t* copy_attrs_to_frame`?

bcarlson-dev (changes requested):

inc/ec_util.h
tuckerpo-cl: `auto` deduces the type from the C-style cast on the RHS of the expression. `reinterpret_cast` isn't needed as it isn't inheritence in a polymorphism sense, only inheritence for data composition. If anything, it should be `static_cast<ec_gas_initial_request_frame_t *>(frame)` which I'm happy to switch to if we don't want C-style casts.

bcarlson-dev: Honestly the main reason for the other casts is to avoid the warnings (since soon it's going to be "treat warnings as errors") so if the current method avoids that then I'm all for it

tuckerpo-cl: Since they're going for `-Werror -Wall` I just switched to C++ casts /shrug

bcarlson-dev (dismissed):
Looks good to me!

bcarlson-dev (approved):

PR #219, opened 2025-03-13
Adds the ability to add/remove vendor IEs via EasyMesh and abstracts webconfig south-bound refresh. 

bcarlson-dev:
- Adds methods, structs, and members to easily add and remove vendor IEs to a BSS
- Abstracts the `webconfig_proto_easymesh_init`, `webconfig_easymesh_encode` and `bus_set_fn` for sending current EasyMesh data model over the south bound interface into a separate method for easier use. (for example, when updating the IEs).

**Depends on OneWifi [PR #206](https://github.com/rdkcentral/OneWifi/pull/206)**

This has been tested with the correct OneWifi version and works when applying this patch and waiting for an action frame to be received (or just sending one to it). This was the easiest way to do asynchronous testing in the current version of the **agent** in my imagination but the core code could be moved else where the reviewer has a better idea.  
[vs_test.patch](https://github.com/user-attachments/files/19237935/vs_test.patch)

@amarnathhullur I am adding you as a reviewer since this deals with `webconfig` and while it does work as expected, appears to mostly be calling the same code and has feature parity with the master branch, I want to make sure I am not missing anything.

### DIFF
--- a/src/dm/dm_bss.cpp
+++ b/src/dm/dm_bss.cpp
@@ -154,6 +154,25 @@ int dm_bss_t::decode(const cJSON *obj, void *parent_id)
     if ((tmp = cJSON_GetObjectItem(obj, "TransmittedBSSID")) != NULL) {
         m_bss_info.transmitted_bssid = cJSON_IsTrue(tmp);
     }
+    
+    const char* vendor_ies = NULL;
+    if ((tmp = cJSON_GetObjectItem(obj, "ExtraVendorIEs")) != NULL && (vendor_ies = cJSON_GetStringValue(tmp)) != NULL) {
+        m_bss_info.vendor_elements_len = strlen(vendor_ies);
+
+        unsigned int element;
+        unsigned int i;
+        for (i = 0; i < sizeof(m_bss_info.vendor_elements); i++) {
+            // Make sure we have two characters for a valid hex number.
+            if (2 * i + 2 > m_bss_info.vendor_elements_len)
+                break;
+            if (sscanf(vendor_ies + 2 * i, "%02x", &element) == 1) {
+                m_bss_info.vendor_elements[i] = (unsigned char)element;
+            } else {
+                break;
+            }
+        }
+        m_bss_info.vendor_elements_len = i;
+    }
 
     return 0;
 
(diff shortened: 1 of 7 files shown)


src/dm/dm_bss.cpp
```diff
@@ -232,6 +251,15 @@ void dm_bss_t::encode(cJSON *obj, bool summary)
     // Add the array to the object
     cJSON_AddItemToObject(obj, "BackhaulAKMsAllowed", backhaul_akmsArray);
 
+    // Add vendor elements (ExtraVendorIEs) as hex string
+    char vendor_ies[2 * m_bss_info.vendor_elements_len + 1] = "";
```
amarnathhullur, line 255: Would be better to set it to '\0' which is more readable

src/dm/dm_bss.cpp
```diff
@@ -232,6 +251,15 @@ void dm_bss_t::encode(cJSON *obj, bool summary)
...
+    if (m_bss_info.vendor_elements_len > 0) {
+        for (unsigned int i = 0; i < m_bss_info.vendor_elements_len; i++) {
+            snprintf(vendor_ies + 2 * i, 3, "%02x", m_bss_info.vendor_elements[i]);
```
amarnathhullur, line 258: why is the 2nd parameter hardcoded to 3,  this should be the remaining size of vendor_ies

src/dm/dm_bss.cpp
```diff
@@ -365,6 +397,67 @@ int dm_bss_t::parse_bss_id_from_key(const char *key, em_bss_id_t *id)
...
+    m_bss_info.vendor_elements_len += vs_ie_len;
+    printf("Successfully added Vendor IE of length %d to BSS\n", vs_ie_len);
+    return true;
+}
+
+void dm_bss_t::remove_vendor_ie(struct ieee80211_vs_ie *vs_ie)
+{
+    size_t vs_ie_len = offsetof(struct ieee80211_vs_ie, vs_oui) + vs_ie->vs_len;
+    if (m_bss_info.vendor_elements_len < vs_ie_len) {
+        // The IE is not present in the BSS, return true since it's technically removed
```
amarnathhullur, line 421: Add a printf statement here

src/dm/dm_bss.cpp
```diff
@@ -365,6 +397,67 @@ int dm_bss_t::parse_bss_id_from_key(const char *key, em_bss_id_t *id)
...
+
+    // Find the IE in the BSS
+    uint8_t* curr_ie_head = m_bss_info.vendor_elements;
+    uint8_t* end = m_bss_info.vendor_elements + m_bss_info.vendor_elements_len;
+    while (curr_ie_head < end) {
+        struct ieee80211_vs_ie* curr_ie = reinterpret_cast<struct ieee80211_vs_ie*>(curr_ie_head);
+        size_t curr_ie_len = offsetof(struct ieee80211_vs_ie, vs_oui) + curr_ie->vs_len;
+
+        // If the current IE is not the same length as the IE we're looking for, skip it
+        if (vs_ie_len != vs_ie_len) {
```
amarnathhullur, line 433: Looks like a typo, this statement cannot become true any time

src/dm/dm_bss.cpp
bcarlson-dev: I can't do that since it won't allow me assign a char to a char array but I can do `{0}` if that is better for you?

src/dm/dm_bss.cpp
bcarlson-dev: Great catch

src/dm/dm_bss.cpp
amarnathhullur: Yeah, that would be better.

bcarlson-dev: Okay it's done

PR #221, opened 2025-03-15
Fix the stability issues in controller and agent

AswiniViswanathan:
Fixed the following issues
1)CTRL DB BSSID table incorrect values  [duplicate entries]
2)Freq band incorrect 
3)Update to CTRL DB not woking for SSID and BSSID configs
4)freq band not updated to DB
5)Autoconfig renew based al mac is empty and added to check for al mac
7)Set channel fail in controller em even when agent is configured 
8)Autoconfig renew fail when happens for both the radio


src/dm/dm_radio_list.cpp
```diff
@@ -196,7 +196,7 @@ int dm_radio_list_t::update_db(db_client_t& db_client, dm_orch_type_t op, void *
     switch (op) {
 		case dm_orch_type_db_insert:
 			ret = insert_row(db_client, key, radio_mac_str, info->enabled,
-						info->media_data.media_type, info->media_data.band, info->media_data.center_freq_index_1, info->media_data.center_freq_index_1,
```
amarnathhullur, line 199: What is the band in media_data which was earlier used?

src/dm/dm_radio_list.cpp
```diff
@@ -261,6 +261,7 @@ int dm_radio_list_t::sync_db(db_client_t& db_client, void *ctx)
         info.enabled = db_client.get_number(ctx, 3);
         info.media_data.media_type = db_client.get_number(ctx, 4);
         info.media_data.band = db_client.get_number(ctx, 5);
+	info.band = (em_freq_band_t) db_client.get_number(ctx, 5);
```
amarnathhullur, line 264: What is the need for this change?

src/em/channel/em_channel.cpp
```diff
@@ -1950,7 +1952,9 @@ void em_channel_t::process_msg(unsigned char *data, unsigned int len)
             break;
 
         case em_msg_type_channel_sel_rsp:
-            handle_channel_sel_rsp(data, len);
+		if (get_service_type() == em_service_type_ctrl) {
```
amarnathhullur, line 1955: Is the understanding that this needs to be processed only in controller?

src/em/config/em_configuration.cpp
```diff
@@ -255,7 +255,7 @@ int em_configuration_t::send_autoconfig_renew_msg()
     tlv = reinterpret_cast<em_tlv_t *> (tmp);
     tlv->type = em_tlv_type_supported_freq_band;
     tlv->len = htons(sizeof(unsigned char));
-    freq_band = static_cast<em_freq_band_t> (get_band() >> 1);
+    freq_band = static_cast<em_freq_band_t> (get_band());
```
amarnathhullur, line 258: This is a significant change as well. Why is the right shift not required now?

src/em/config/em_configuration.cpp
```diff
@@ -838,6 +840,19 @@ void em_configuration_t::handle_ap_vendor_operational_bss(unsigned char *value,
 		for (j = 0; j < radio->bss_num; j++) {
 			dm_easy_mesh_t::macbytes_to_string(bss->bssid, bss_mac_str);
 			printf("%s:%d: BSSID=%s	 haul type=%d\n", __func__, __LINE__, bss_mac_str, bss->haultype);
+			dm_bss = dm->get_bss(radio->ruid, bss->bssid);
+			if (dm_bss == NULL) {
+				dm_bss = &dm->m_bss[dm->m_num_bss];
+				dm->set_num_bss(dm->get_num_bss() + 1);
+			}
+			// fill up id firs
```
amarnathhullur, line 848: typo in first

src/em/config/em_configuration.cpp
```diff
@@ -1609,7 +1624,19 @@ int em_configuration_t::handle_topology_response(unsigned char *buff, unsigned i
         
     tlv =  reinterpret_cast<em_tlv_t *> (buff + sizeof(em_raw_hdr_t) + sizeof(em_cmdu_t));
     tmp_len = len - static_cast<unsigned int> (sizeof(em_raw_hdr_t) + sizeof(em_cmdu_t));
-        
+
+	while ((tlv->type != em_tlv_type_eom) && (tmp_len > 0)) {
```
amarnathhullur, line 1628: Looks like 1628 to 1637 is same as that of 1640 to 1651, why is this while loop repeated.

amarnathhullur: Need more comments

src/em/config/em_configuration.cpp
```diff
@@ -2028,7 +2045,7 @@ unsigned short em_configuration_t::create_m2_msg(unsigned char *buff, em_haul_ty
     attr->id = htons(attr_id_rf_bands);
     size = 1;
     attr->len = htons(size);
-    rf_band = get_band();
+    rf_band = static_cast<em_freq_band_t> (1 << get_band());
```
amarnathhullur, line 2048: Not sure about this change. Need discussion

src/em/config/em_configuration.cpp
```diff
@@ -2892,10 +2909,25 @@ int em_configuration_t::handle_wsc_m1(unsigned char *buff, unsigned int len)
...
     dm = get_data_model();
 	memset(&dev_info, 0, sizeof(em_device_info_t));
-    
+
+	for (i = 0; i < dm->m_num_radios; i++) {
+		radio = dm->get_radio(i);
+		if (memcmp(radio->get_radio_interface_mac(), get_radio_interface_mac(), sizeof(mac_address_t)) != 0) {
+			continue;
+		}
+	   found++;
```
amarnathhullur, line 2923: indentation incorrect

src/orch/em_orch_ctrl.cpp
```diff
@@ -110,6 +109,13 @@ bool em_orch_ctrl_t::is_em_ready_for_orch_fini(em_cmd_t *pcmd, em_t *em)
             break;
         
         case em_cmd_type_set_channel:
+	    if (em->get_state() == em_state_ctrl_channel_selected) {
+		return true;
+             }
+	    else if (em->get_state() == em_state_ctrl_configured) {
+                 return true;
+	    }
```
amarnathhullur, line 117: you can add an || instead of if, elseif as you are returning same value

src/orch/em_orch_ctrl.cpp
```diff
@@ -211,10 +217,18 @@ bool em_orch_ctrl_t::is_em_ready_for_orch_exec(em_cmd_t *pcmd, em_t *em)
...
             break;
 
+	case em_cmd_type_set_channel:
+	   if (em->get_state() == em_state_ctrl_configured) {
+                return true;
+            } else if (em->get_state() == em_state_ctrl_misconfigured) {
+                return true;
+            } else if( em->get_state() == em_state_ctrl_channel_selected) {
+		return true;
+	    }
```
amarnathhullur, line 227: you can add an || instead of if, elseif as you are returning same value

src/orch/em_orch_ctrl.cpp
```diff
@@ -392,11 +406,11 @@ unsigned int em_orch_ctrl_t::build_candidates(em_cmd_t *pcmd)
     while (em != NULL) {
         switch (pcmd->m_type) {
             case em_cmd_type_set_ssid:
-                for (i = 0; i < pcmd->get_num_network_ssid(); i++) {
-                    if (em->is_set_ssid_candidate(pcmd->get_network_ssid(i))) {
-                        queue_push(pcmd->m_em_candidates, em);
-                        count++;
-                    }
+		if (em->is_al_interface_em() == false) {
```
amarnathhullur, line 409: The for loop is removed, significant change.

src/orch/em_orch_ctrl.cpp
```diff
@@ -392,11 +406,11 @@ unsigned int em_orch_ctrl_t::build_candidates(em_cmd_t *pcmd)
...
+			dm_easy_mesh_t::macbytes_to_string(em->get_radio_interface_mac(), mac_str);
+			printf("%s:%d Set SSSID : %s push to queue \n", __func__, __LINE__,mac_str);
```
amarnathhullur, line 411: typo in SSID

amarnathhullur (commented):
Lets have a discussion on these changes as some changes need more clarification.

src/em/config/em_configuration.cpp
bcarlson-dev: Can you add some documentation to the `em_freq_band_t` struct detailing this discrepancy and that the same struct is used for both versions so a conversion is necessary?

bcarlson-dev: Similarly above`get_band` or the `m_band` variable it might be useful to have documentation detailing wether that is (without conversion) the RF band version or the EM TLV band version.

src/em/config/em_configuration.cpp
bcarlson-dev: I think this is fine but I think there should be some documentation here to avoid the understandable confusion here on why an `em_freq_band_t` is being converted to an `em_freq_band_t`. 

Maybe something as simple as "Converting from EM TLV Freq Band representation to RF Freq Band Representation. See `em_freq_band_t` for more information"

src/dm/dm_radio_list.cpp
amarnathhullur: Should the webconfig translator and the agent needs to change to update media_data.band instead of using info.band?

src/orch/em_orch_ctrl.cpp
amarnathhullur: Ok, how will this be applied for all EM? Is this function called for all EM?

src/orch/em_orch_ctrl.cpp
```diff
@@ -455,6 +468,14 @@ unsigned int em_orch_ctrl_t::build_candidates(em_cmd_t *pcmd)
                 break;
             
             case em_cmd_type_set_channel:
+                dm = pcmd->get_data_model();
```
amarnathhullur, line 471: How was set_channel working earlier without the below change?

src/em/config/em_configuration.cpp
AswiniViswanathan: typedef enum {
    em_freq_band_24,    //IEEE-1905-1-2013 table 6-23
    em_freq_band_5,
    em_freq_band_60,
    em_freq_band_unknown
} em_freq_band_t;
 In the header file, the spec is mentioned.

bcarlson-dev: Yes, however, based on that reference, someone would think that it is only used for `AutoconfigFreqBand` and maybe `SupportedFreqBand` (6.4.17). It is used other places though as a general enum representing 2.4, 5, and 60 GHz (as we see with the RF band), so I think a more detailed documentation should be added saying that this enum is used in places outside of 1905 TLVs at which point integer conversion may be necessary. It might even be a good idea to call out the RF Band specifically as a conversion that occurs based on its different representation in the IEEE spec.

It doesn't hurt, especially when you are using the same enum across two different specs, with two different representations.

src/em/config/em_configuration.cpp
```diff
@@ -2891,11 +2908,26 @@ int em_configuration_t::handle_wsc_m1(unsigned char *buff, unsigned int len)
...
 
-    dm = get_data_model();
+	dm = get_data_model();
 	memset(&dev_info, 0, sizeof(em_device_info_t));
-    
+
+	for (i = 0; i < dm->m_num_radios; i++) {
+		radio = dm->get_radio(i);
+		if (memcmp(radio->get_radio_interface_mac(), get_radio_interface_mac(), sizeof(mac_address_t)) != 0) {
+				continue;
```
amarnathhullur, line 2921: indentation incorrect, two tabs for continue

src/agent/em_agent.cpp
```diff
@@ -287,9 +287,7 @@ void em_agent_t::handle_autoconfig_renew(em_bus_event_t *evt)
     em_cmd_t *pcmd[EM_MAX_CMD] = {NULL};
     unsigned int num;
 
-    if (m_orch->is_cmd_type_in_progress(evt->type) == true) {
```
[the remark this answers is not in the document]
AswiniViswanathan: Reverted since its not required

src/dm/dm_radio_list.cpp
AswiniViswanathan: I will discuss with soumya and get the information on media_data struct and update it the right way, but for M2 we have to updated the band variable.

src/em/config/em_configuration.cpp
AswiniViswanathan: sure Ben, I will add few more details in comments

src/em/config/em_configuration.cpp
AswiniViswanathan: As discussed the loop is required since we have to process one TLV after another in specific order only

src/em/config/em_configuration.cpp
AswiniViswanathan: Got it. Will add them

src/em/config/em_configuration.cpp
AswiniViswanathan: Addressed.

amarnathhullur (approved):

src/em/config/em_configuration.cpp
amarnathhullur: Merging this after indicating Aswini to update the comment in the next commit.

amarnathhullur (approved):

PR #230, opened 2025-03-19
make unwrap_wrapped_attrib generic

tuckerpo-cl:


src/em/prov/easyconnect/ec_util.cpp
```diff
@@ -184,24 +189,23 @@ std::pair<uint8_t*, uint16_t> ec_util::unwrap_wrapped_attrib(ec_attribute_t* wra
     uint8_t* unwrap_attribs = new uint8_t[wrapped_len]();
     int result = -1;
     if (uses_aad) {
-        if (frame == NULL) {
```
bcarlson-dev, line 187: Why was this check removed?

bcarlson-dev (changes requested):

src/em/prov/easyconnect/ec_util.cpp
tuckerpo-cl: `siv_decrypt` can take a `nullptr` and safely fails AFAIK but I can add the null check back in

bcarlson-dev: Yeah can you do that just so there is that more descriptive error message

bcarlson-dev (approved):

PR #234, opened 2025-03-21
Adds callbacks for retreiving bSTA info / 1905 layer info

tuckerpo-cl:
Implement DPP Configuration Request Object creation, creating bSTAList object on Enrollee

Begins parsing of DPP Configuration Frame on Controller

Also plugs memory leaks in Enrollee.

Configuration Request frame parsing / Configuration Response frame handling still TODO in Configurator, need callbacks first.

See EasyConnect 6.4.3.1 for auditing Configurator work, EasyMesh 5.3.2 Table 5 for auditing Enrollee `get_bsta_info` callback.

Note: `akm` is still hard-coded as there's nothing (currently) in the data model to expose this nor map it to a radio.

### DIFF
--- a/inc/ec_configurator.h
+++ b/inc/ec_configurator.h
@@ -57,13 +57,13 @@ using get_backhaul_sta_info_func = std::function<cJSON*(ec_connection_context_t 
 
 /**
  * @brief Creates a DPP Configuration Response object for the 1905.1 interface.
- *
+ * @return cJSON * on success, nullptr otherwise.
  */
 using get_1905_info_func = std::function<cJSON*(ec_connection_context_t *)>;
 
 /**
  * @brief Used to determine if an additional AP can be on-boarded or not.
- *
+ * @return True if additional APs can be on-boraded into the mesh, false otherwise.
  */
 using can_onboard_additional_aps_func = std::function<bool(void)>;
 
@@ -176,6 +176,16 @@ public:
      */
     virtual bool  process_proxy_encap_dpp_msg(em_encap_dpp_t *encap_tlv, uint16_t encap_tlv_len, em_dpp_chirp_value_t *chirp_tlv, uint16_t chirp_tlv_len) = 0;
 
+    /**
+     * @brief Handle a proxied encapsulated DPP Configuration Request frame.
+     * 
+     * @param encap_frame The DPP Configuration Request frame from an Enrollee.
+     * @param encap_frame_len The length of the DPP Configuration Request frame.
+     * @param dest_mac The source MAC of this DPP Configuration Request frame (Enrollee).
+     * @return true on success, otherwise false.
+     * 
+     * @note: overridden by subclass.
+     */
     virtual bool handle_proxied_dpp_configuration_request(uint8_t *encap_frame, uint16_t encap_frame_len, uint8_t dest_mac[ETH_ALEN]) {
         return true;
     }
--- a/inc/ec_configurator.h
+++ b/inc/ec_configurator.h
@@ -48,6 +48,25 @@ using send_act_frame_func = std::function<bool(uint8_t*, uint8_t *, size_t, unsi
 */
 using toggle_cce_func = std::function<bool(bool)>;
 
+/**
+ * @brief Creates a DPP Configuration Response object for the backhaul STA interface.
+ * @param conn_ctx Optional connection context (not needed for Enrollee, needed for Configurator) -- pass nullptr if not needed.
+ * @return cJSON * on success, nullptr otherwise
+ */
+using get_backhaul_sta_info_func = std::function<cJSON*(ec_connection_context_t *)>;
+
+/**
+ * @brief Creates a DPP Configuration Response object for the 1905.1 interface.
+ *
+ */
+using get_1905_info_func = std::function<cJSON*(ec_connection_context_t *)>;
+
+/**
+ * @brief Used to determine if an additional AP can be on-boarded or not.
+ *
+ */
+using can_onboard_additional_aps_func = std::function<bool(void)>;
+
 class ec_configurator_t {
 public:
     /**
@@ -58,7 +77,8 @@ public:
      */
     // TODO: Add send_gas_frame functions
     ec_configurator_t(std::string mac_addr, send_chirp_func send_chirp_notification, send_encap_dpp_func send_prox_encap_dpp_msg, 
-                        send_act_frame_func send_action_frame);
+                        send_act_frame_func send_action_frame, get_backhaul_sta_info_func backhaul_sta_info_func, get_1905_info_func ieee1905_info_func,
+                        can_onboard_additional_aps_func can_onboard_func);
     virtual ~ec_configurator_t(); // Destructor
 
     /**
@@ -156,6 +176,10 @@ public:
      */
     virtual bool  process_proxy_encap_dpp_msg(em_encap_dpp_t *encap_tlv, uint16_t encap_tlv_len, em_dpp_chirp_value_t *chirp_tlv, uint16_t chirp_tlv_len) = 0;
 
+    virtual bool handle_proxied_dpp_configuration_request(uint8_t *encap_frame, uint16_t encap_frame_len, uint8_t dest_mac[ETH_ALEN]) {
+        return true;
+    }
+
     inline std::string get_mac_addr() { return m_mac_addr; };
 
     // Disable copy construction and assignment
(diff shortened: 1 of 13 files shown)


src/em/em.cpp
```diff
@@ -1078,14 +1078,20 @@ em_t::em_t(em_interface_t *ruid, em_freq_band_t band, dm_easy_mesh_t *dm, em_mgr
...
         mac_address,
         std::bind(&em_t::send_chirp_notif_msg, this, std::placeholders::_1, std::placeholders::_2),
         std::bind(&em_t::send_prox_encap_dpp_msg, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, std::placeholders::_4),
         std::bind(&em_mgr_t::send_action_frame, mgr, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, std::placeholders::_4), 
-        get_service_type() == em_service_type_ctrl
+        service_type == em_service_type_agent
+            ? std::bind(&em_t::create_enrollee_bsta_list, this, std::placeholders::_1)
+            : static_cast<std::function<cJSON*(ec_connection_context_t *)>>(nullptr),
+        nullptr,
+        nullptr,
```
bcarlson-dev, line 1093: Can you add some inline comments explaining the `nullptr`s a bit, at least mentioning that they are temporary? Optionally you *could* do a dummy lambda like the current ["Toggle CCE" function](https://github.com/rdkcentral/unified-wifi-mesh/blob/main/src/em/config/em_configuration.cpp#L3141) but that would add more to this constructor than is nescessary right now

bcarlson-dev (changes requested):

bcarlson-dev (changes requested):
There are two missing frees but about half of the comments are documentation related and about half of them are **optional** ways to potentially clean up code (although as always the balance between readability and cleanliness is up to you). Besides those it LGTM!

src/em/prov/em_provisioning.cpp
```diff
@@ -543,6 +545,121 @@ int em_provisioning_t::handle_proxy_encap_dpp(uint8_t *buff, unsigned int len)
...
+    cJSON *b_sta_list_arr = cJSON_CreateArray();
+    if (!b_sta_list_arr) {
+        printf("%s:%d: Could not allocate for bSTAList array!\n", __func__, __LINE__);
+        cJSON_Delete(bsta_list_obj);
+        return nullptr;
+    }
+
+    if (!cJSON_AddStringToObject(bsta_list_obj, "netRole", "mapBackhaulSta")) {
+        printf("%s:%d: Could not add netRole to bSTAList object!\n", __func__, __LINE__);
+        cJSON_Delete(b_sta_list_arr);
```
bcarlson-dev, line 572: I think you need to `cJSON_Delete(bsta_list_obj)`.

Based on [this example](https://github.com/DaveGamble/cJSON/blob/master/tests/readme_examples.c#L135) though, I think you can change things around so that the `cJSON_CreateArray` is actually a `cJSON_AddArrayToObject` at which point all of your `cJSON_Delete`s can just be freeing the root `bsta_list_obj`

src/em/prov/em_provisioning.cpp
```diff
@@ -543,6 +545,121 @@ int em_provisioning_t::handle_proxy_encap_dpp(uint8_t *buff, unsigned int len)
...
+        return nullptr;
+    }
+
+    cJSON *radio_list_arr = cJSON_AddArrayToObject(bsta_list_obj, "RadioList");
+    if (!radio_list_arr) {
+        printf("%s:%d: Could not add RadioList array to bSTAList object!\n", __func__, __LINE__);
+        cJSON_Delete(bsta_list_obj);
+        return nullptr;
+    }
```
bcarlson-dev, line 602: If `cJSON_AddArrayToObject` is used, this section can be removed

src/em/prov/em_provisioning.cpp
```diff
@@ -543,6 +545,121 @@ int em_provisioning_t::handle_proxy_encap_dpp(uint8_t *buff, unsigned int len)
...
+            return nullptr;
+        }
+
+        if (!cJSON_AddStringToObject(
+                radioListObj, "RUID",
+                util::mac_to_string(radio->m_radio_info.id.ruid, "").c_str())) {
+            printf("%s:%d: Could not add RUID to RadioList object!\n", __func__, __LINE__);
+            cJSON_Delete(bsta_list_obj);
+            return nullptr;
+        }
```
bcarlson-dev, line 622: Based on the example previously given this seems *fine* but its weird that it doesn't free the `radioListObj`.
I can't tell if that's an oversight in the example or intentional. Either way, you do it later in the code so it might be a good idea to just do it here.

```suggestion
        if (!cJSON_AddStringToObject(
                radioListObj, "RUID",
                util::mac_to_string(radio->m_radio_info.id.ruid, "").c_str())) {
            printf("%s:%d: Could not add RUID to RadioList object!\n", __func__, __LINE__);
            cJSON_Delete(bsta_list_obj);
            cJSON_Delete(radioListObj);
            return nullptr;
        }
```

src/em/prov/em_provisioning.cpp
```diff
@@ -543,6 +545,121 @@ int em_provisioning_t::handle_proxy_encap_dpp(uint8_t *buff, unsigned int len)
...
+    }
+
+    if (!cJSON_AddStringToObject(bsta_list_obj, "channelList", channelList.c_str())) {
+        printf("%s:%d: Could not add channelList to bSTAList object!\n", __func__, __LINE__);
+        cJSON_Delete(bsta_list_obj);
+        return nullptr;
+    }
+    return b_sta_list_arr;
```
bcarlson-dev, line 660: Isn't the object supposed to be returned here?
```suggestion
    return bsta_list_obj;
```

inc/ec_configurator.h
```diff
@@ -48,6 +48,25 @@ using send_act_frame_func = std::function<bool(uint8_t*, uint8_t *, size_t, unsi
...
+ * @param conn_ctx Optional connection context (not needed for Enrollee, needed for Configurator) -- pass nullptr if not needed.
+ * @return cJSON * on success, nullptr otherwise
+ */
+using get_backhaul_sta_info_func = std::function<cJSON*(ec_connection_context_t *)>;
+
+/**
+ * @brief Creates a DPP Configuration Response object for the 1905.1 interface.
+ *
+ */
+using get_1905_info_func = std::function<cJSON*(ec_connection_context_t *)>;
```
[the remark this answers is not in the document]
tuckerpo-cl: I'll add `@param` here, but `can_onboard_additional_aps_func` doesn't take any arguments.

src/em/prov/em_provisioning.cpp
tuckerpo-cl: `bSTAList` is the "parent" JSON object here with `bsta_list_obj` inside of it. 

image


`bSTAList` object consists of `akm`, `channelList`, `RadioList` arr, `netRole` and `bSTA_Maximum_Link`

At the callsite in `ec_enrollee.cpp:675` the array is returned and added with `"bSTAList"` as the key to the old DPP Configuration Request Object

src/em/em.cpp
```diff
@@ -1078,14 +1078,20 @@ em_t::em_t(em_interface_t *ruid, em_freq_band_t band, dm_easy_mesh_t *dm, em_mgr
...
     std::string mac_address = util::mac_to_string(get_peer_mac());
     m_ec_manager = std::unique_ptr<ec_manager_t>(new ec_manager_t(
         mac_address,
         std::bind(&em_t::send_chirp_notif_msg, this, std::placeholders::_1, std::placeholders::_2),
         std::bind(&em_t::send_prox_encap_dpp_msg, this, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, std::placeholders::_4),
         std::bind(&em_mgr_t::send_action_frame, mgr, std::placeholders::_1, std::placeholders::_2, std::placeholders::_3, std::placeholders::_4), 
-        get_service_type() == em_service_type_ctrl
+        service_type == em_service_type_agent
+            ? std::bind(&em_t::create_enrollee_bsta_list, this, std::placeholders::_1)
+            : static_cast<std::function<cJSON*(ec_connection_context_t *)>>(nullptr),
```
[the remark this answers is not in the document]
tuckerpo-cl: Yeah this is really ugly -- once the Configurator's version of `get_bsta_info_func` is written it'll be bound here and the ternary will look a _little_ less ridiculous. But yeah since we define `get_backhaul_sta_info_func` as a type we can swap that into the `static_cast<T>`, good call

src/em/prov/em_provisioning.cpp
bcarlson-dev: Got it! I totally missed that. One question. Since this is in `em_t`? Doesn't this only give the info for one BSS? Shouldn't this implementation be located in the `em_agent_t` and passed up via the `em_agent_t` -> `em_mgr_t` (then bound in the `em_t` initializer) similar to `send_action_frame`?

bcarlson-dev (approved):

PR #245, opened 2025-03-27
Added Unit Test GitHub Action Workflow

bcarlson-dev:
Adds an additional GitHub Action workflow to test whatever Unit Tests are present in the `tests` folder. 

This is currently just for unit tests (not end-to-end integration tests using OneWifi, the Agent, and the Controller) and these tests have limited coverage for critical atomic methods. Due to my current work, these only relate to EasyConnect and cryptography but the additional unit-testing files can be added to the "tests" folder in the future and this GitHub action will not need modification.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,0 +1,76 @@
+name: Tests
+
+on:
+  push:
+    branches: [ "main" ]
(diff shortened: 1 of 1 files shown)


amarnathhullur (approved):
Looks good to me

PR #246, opened 2025-03-27
Refactored EasyConnect code to better handle multiple enrollees

bcarlson-dev:


src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -376,71 +390,70 @@ bool ec_ctrl_configurator_t::handle_proxied_dpp_configuration_request(uint8_t *e
...
         // Add ppKey
 
-        cJSON *ppKeyObj = ec_crypto::create_ppkey_object(m_p_ctx);
+        cJSON *ppKeyObj = ec_crypto::create_ppkey_object(*conn_ctx);
         cJSON_AddItemToObject(cred, "ppKey", ppKeyObj);
 
     }
+    // Allocates new memory for the strings
+    char* ieee1905_config_obj_str = cJSON_Print(ieee1905_config_obj);
+    char* bsta_config_object_str = cJSON_Print(bsta_config_object);
```
tuckerpo-cl, line 428: These `char *` need to be `free`'d once done being used since `cJSON_Print` heap allocs + returns a copy of the object in string form

Alternatively, move to `cjson_utils.h`:

```c++
cjson_utils::stringify(cJSON *blob)
{
    char *s = cJSON_Print(blob);
    std::string str(s);
    free(s);
    return str;
}
```

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -213,21 +228,24 @@ bool ec_ctrl_configurator_t::handle_proxied_dpp_configuration_request(uint8_t *e
...
         response_frame->resp_len = static_cast<uint16_t>(attribs_len);
         auto [proxy_encap_frame, proxy_encap_frame_len] = ec_util::create_encap_dpp_tlv(true, src_mac, ec_frame_type_easymesh, reinterpret_cast<uint8_t*>(response_frame), sizeof(*response_frame) + attribs_len);
         if (proxy_encap_frame == nullptr || proxy_encap_frame_len == 0) {
             printf("%s:%d: Could not create Proxied Encap DPP TLV!\n", __func__, __LINE__);
             free(response_frame);
-            free(attribs);
             return false;
         }
-        printf("%s:%d: Sending DPP Configuration Response frame for Enrollee '" MACSTRFMT "' over 1905 with DPP status code %s\n", __func__, __LINE__, MAC2STR(src_mac), ec_util::status_code_to_string(static_cast<ec_status_code_t>(response_frame->status_code)).c_str());
+        std::string status_code_str =  ec_util::status_code_to_string(static_cast<ec_status_code_t>(response_frame->status_code));
```
tuckerpo-cl, line 240: Realizing now that I never set `response_frame->status_code = <thing>` outside of the `if (cannot_onboard_more)` case -- could you just set it to `DPP_STATUS_OK` here?

tuckerpo-cl (changes requested):

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
bcarlson-dev: Good catch and good idea

PR #247, opened 2025-03-27
ec_enrollee: create DPP Configuration result frames

tuckerpo-cl:
Configuration Result frame and Configuration Connection Status Result frame

Necessary for Enrollee configueration response handling

### DIFF
--- a/src/em/prov/easyconnect/ec_enrollee.cpp
+++ b/src/em/prov/easyconnect/ec_enrollee.cpp
@@ -821,7 +821,6 @@ std::pair<uint8_t *, size_t> ec_enrollee_t::create_config_request()
     cJSON *bsta_info = m_get_bsta_info(nullptr);
     ASSERT_NOT_NULL_FREE(bsta_info, {}, m_eph_ctx().e_nonce, "%s:%d: bSTA info is nullptr!\n", __func__, __LINE__);
     cJSON_AddItemToObject(dpp_config_request_obj, "bSTAList", bsta_info);
-    size_t config_obj_len = cjson_utils::get_cjson_blob_size(dpp_config_request_obj);
 
     // XXX: Dialog token can be thought of as a session key between Enrollee and Configurator regarding configuration
     // From specs (EasyMesh, EasyConnect, 802.11), it seems this is just arbitrarily chosen (1 byte), but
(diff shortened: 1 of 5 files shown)


src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -888,3 +888,94 @@ std::pair<uint8_t *, size_t> ec_enrollee_t::create_config_result(ec_status_code_
...
+    attribs =
+        ec_util::add_wrapped_data_attr(frame, attribs, &attribs_len, true, m_eph_ctx.ke, [&]() {
+            size_t wrapped_len = 0;
+            uint8_t *wrapped_attrs =
+                ec_util::add_attrib(nullptr, &wrapped_len, ec_attrib_id_enrollee_nonce,
+                                    m_p_ctx.nonce_len, m_eph_ctx.e_nonce);
+            wrapped_attrs = ec_util::add_attrib(
+                wrapped_attrs, &wrapped_len, ec_attrib_id_conn_status,
+                static_cast<uint16_t>(cjson_utils::get_cjson_blob_size(connection_status_object)),
+                reinterpret_cast<uint8_t *>(connection_status_object));
```
bcarlson-dev, line 916: We need to convert `connection_status_object` to a string before adding it as an attribute.

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -888,3 +888,94 @@ std::pair<uint8_t *, size_t> ec_enrollee_t::create_config_result(ec_status_code_
...
+            return std::make_pair(wrapped_attrs, wrapped_len);
+        });
+
+    if (!(frame = ec_util::copy_attrs_to_frame(frame, attribs, attribs_len))) {
+        printf("%s:%d: Failed to copy attributes to DPP Connection Status Result frame\n", __func__,
+               __LINE__);
+        free(attribs);
+        free(frame);
+        free(connection_status_object);
```
bcarlson-dev, line 925: `connection_status_object` needs to be freed with `cJSON_delete`. This can also be performed earlier in the method (line 903) to avoid the memory leak in the next comment

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -888,3 +888,94 @@ std::pair<uint8_t *, size_t> ec_enrollee_t::create_config_result(ec_status_code_
...
+        return {};
+    }
+    return std::make_pair(reinterpret_cast<uint8_t *>(frame), EC_FRAME_BASE_SIZE + attribs_len);
```
bcarlson-dev, line 928: `attribs` needs to be freed here (and `connection_status_object` needs to be `cJSON_delete`d unless you do it sooner in the function)

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -888,3 +888,94 @@ std::pair<uint8_t *, size_t> ec_enrollee_t::create_config_result(ec_status_code_
...
+            first_group = false;
+
+            channel_list += std::to_string(opclass) + "/";
+
+            for (size_t i = 0; i < channels.size(); i++) {
+                if (i > 0)
+                    channel_list += ",";
+                channel_list += std::to_string(channels[i]);
+            }
+        }
```
bcarlson-dev, line 977: Can you abstract this into a general `ec_util` function since this will be useful in creating a DPPURI?

bcarlson-dev (changes requested):

src/em/prov/easyconnect/ec_enrollee.cpp
bcarlson-dev: I just merged my PR which shouldn't be a big disruption to you since you are only in the enrollee for this PR but now the code has the `stringify` method and a new `add_attrib` overload for `std::string`

tuckerpo-cl: Welp, then every wrapped JSON object in `ec_ctrl_configurator` and `ec_enrollee` will need to be `stingify`-ied, they're currently all just `reinterpret_cast`'d to `uint8_t*` -- I can take care of that in another PR, but will change here of course

tuckerpo-cl: You've already taken care of this is `ec_ctrl_configurator`, thanks. I found another spot in `ec_enrollee` in addition to this.

bcarlson-dev (approved):

PR #254, opened 2025-03-29
Implement DPP toggle_cce 

apahl-cl:
As described in EasyMesh R6, Section 5.3.1, proxy agents must be able to advertise the CCE IE in their Beacon and Probe Response frames. The main addition of this PR is `em_t::toggle_cce(bool enable)`, which adds or removes DPP IEs to BSSs to advertise the CCE IE.

Currently, the IEs are added to every BSS found in `em_t::m_data_model`. 

I considered splitting `toggle_cce` into 2 separate functions (`enable_cce` and `disable_cce`), but I decided to stick with `toggle_cce` as it integrates more smoothly into existing code. 

Requesting review from @bcarlson-dev.

### DIFF
--- a/inc/ec_manager.h
+++ b/inc/ec_manager.h
@@ -4,6 +4,7 @@
 #include "ec_configurator.h"
 #include "ec_pa_configurator.h"
 #include "ec_enrollee.h"
+#include "ieee80211.h"
 
 #include <memory>
 #include <functional>
@@ -27,7 +28,7 @@ public:
      * 
      */
     ec_manager_t(std::string mac_addr, send_chirp_func send_chirp, send_encap_dpp_func send_encap_dpp, send_act_frame_func send_action_frame, 
-        get_backhaul_sta_info_func get_bsta_info, get_1905_info_func get_1905_info, can_onboard_additional_aps_func can_onboard, bool m_is_controller);
+        get_backhaul_sta_info_func get_bsta_info, get_1905_info_func get_1905_info, can_onboard_additional_aps_func can_onboard, toggle_cce_func toggle_cce, bool m_is_controller);
     ~ec_manager_t();
 
     /**
@@ -76,6 +77,7 @@ public:
      * 
      * @param enable Whether to enable or disable CCE presence
      * @return bool true if successful, false otherwise
+     * @note If the operation fails, all CCE IEs are removed before the function exits
      */
     inline bool pa_cfg_toggle_cce(bool enable) {
         if (!m_is_controller || m_configurator == nullptr) {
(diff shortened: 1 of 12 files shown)


inc/em_mgr.h
```diff
@@ -61,6 +61,18 @@ class em_mgr_t {
...
+    /**
+     * @brief Refresh the OneWifi subdoc with current information + provided data and send to OneWifi
+     * 
+     * @param log_name [in] The string to use when logging
+     * @param type [in] The subdoc type
+     * @return int 0 if encode fails, -1 if send fails, 1 if both succeed.
+     */
+    virtual int refresh_onewifi_subdoc(const char *log_name, const webconfig_subdoc_type_t type) {
+        printf("refresh_onewifi_subdoc not implemented\n");
+        return 0;
```
bcarlson-dev, line 73: Can we add a new return value here (`-2`?) and in the docstring for this default situation where it is not implemented?

This is important because both the controller and the agent inheirit from the manager but only the agent can (since it's the only one with direct access to OneWifi) implement this. So it's very possible that someone could accidentially call this on the controllers side which would be very improper.

apahl-cl:
Implemented Ben's feedback. Here are all the changes since his review:
- Moved `CCE_IE` into `dm_bss.h`
- Add `toggle_cce` into the `ec_manager_t` constructor
- Remove `toggle_cce` from `ec_manager_t::upgrade_to_onboarded_proxy_agent` function parameters
- Add `toggle_cce` into the `ec_pa_configurator` constructor
- Removed `m_toggle_cce` from `ec_configurator.h` as `ec_pa_configurator` is the only subclass that should be able to toggle CCE
- Add a fourth exit code to `em_agent_t::refresh_onewifi_subdoc`: `-2` means the function is unimplemented
- Use MACSTRFMT as suggested for printing BSSIDs
- Add basic error handling on vendor IE refresh in `em_t::toggle_cce`

src/em/em.cpp
```diff
@@ -635,6 +635,50 @@ bool em_t::is_matching_freq_band(em_freq_band_t *band)
...
+            em_interface_t* bssid = &bss_info->bssid;
+
+            bss->remove_vendor_ie(&dm_bss_t::CCE_IE);
+            printf("Removed DPP IE from BSS '%02x\n", bssid->mac[0]);
+        }
+    }
+
+    // Refresh OneWifi
+    webconfig_subdoc_type_t vap_type = dm_easy_mesh_t::get_subdoc_vap_type_for_freq(get_band());
+    m_mgr->refresh_onewifi_subdoc("'Vendor IE Refresh'", vap_type);
```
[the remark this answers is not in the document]
apahl-cl: My error handling is a bit rudimentary for this, I just check for an unsuccessful return value, print the return value if it's an error, and return false. But I think that's sufficient for `toggle_cce` since it just returns a bool, unless we want to implement finer-grained error return from `toggle_cce`.

inc/ec_manager.h
```diff
@@ -95,10 +97,11 @@ class ec_manager_t {
      * 
      * @param toggle_cce The function to call to toggle CCE presence
      * @return bool true if successful, false otherwise
+     * @note If the operation fails, all CCE IEs are removed before the function exits
```
bcarlson-dev, line 100: Can you remove the `@param` from the docstring now that it's no longer needed?

bcarlson-dev (changes requested):
Since that small docstring change is addressed it LGTM!

inc/ec_manager.h
apahl-cl: Of course, done!

bcarlson-dev (approved):

PR #259, opened 2025-04-01
Fixed major + minor bugs and disabled autoconfig search on startup when Enrollee is DPP onboarding

bcarlson-dev:


src/em/config/em_configuration.cpp
```diff
@@ -3698,6 +3698,12 @@ void em_configuration_t::process_msg(unsigned char *data, unsigned int len)
 
 void em_configuration_t::handle_state_config_none()
 {
+    bool is_dpp_onboarding = get_ec_mgr().is_enrollee_onboarding();
+    if (is_dpp_onboarding) {
+        // Enrollee is in onboarding state, so we should not send autoconfig search message
+        return;
+    }
+
```
tuckerpo-cl, line 3706: Is `m_enrollee->start_onboarding(do_reconfig, boot_data);` called prior to the `em_t` state machine pumping?

If not, `m_is_e_onboarding` should probably default to `true`

We also need `m_is_e_onboarding` to be `true` once Enrollee configuration is finished, but I imagine that will come in a later PR.

tuckerpo-cl (approved):

src/em/config/em_configuration.cpp
tuckerpo-cl: Come to think of it, should `ec_enrollee` hold a reference to the manager? `m_enrollee->set_mgr(this);` -- so Enrollee can signal that configuration has finished `this->mgr.set_enrollee_onboarding(false);` or something similar

bcarlson-dev: Yeah it is and I have tested that it works. I made this a separate PR because I noticed that behavior.

I was also thinking about how we will probably need to manipulate that function a bit more since we need to run autoconf but I figured that might as well be in it's own PR

PR #268, opened 2025-04-03
Read/Write Keypair PEM + DPP bootstrapping PEM saving

bcarlson-dev:
- Adds the new `em_crypto_t::write_keypair_to_pem` and `em_crypto_t::read_keypair_from_pem` functions. 
- Implements the saving and reading of local bootstrapping PEM files for private key information.

### DIFF
--- a/inc/ec_base.h
+++ b/inc/ec_base.h
@@ -44,6 +44,7 @@ extern "C"
 // START: Hardcoded EasyConnect values
 #define DPP_VERSION 0x02
 #define DPP_URI_JSON_PATH "/nvram/DPPURI.json"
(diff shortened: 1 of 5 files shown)


tuckerpo-cl (approved):
`write/read` seem to be duplicated although I assume that's on both sides of the OpenSSL version check boundary, so no problem if intentional

bcarlson-dev:
> `write/read` seem to be duplicated although I assume that's on both sides of the OpenSSL version check boundary, so no problem if intentional

Yeah if you look at it a bit more they use different functions for the reading and writing since they are for different keys (the SSL_KEY just obscures that fact)

PR #272, opened 2025-04-05
ec_ctrl_configurator: break up handle_proxied_dpp_configuration_request

tuckerpo-cl:
Move Encap DPP TLV creation into create_config_response

Add complete_config_obj based on Configuration object type

I think `complete_config_obj` is a terrible name but couldn't think of anything better -- open to suggestions

Additionally adds a dummy handler for `can_onboard_additional_aps`:

 (EasyConnect 6.4.3.1 "If the Configurator does not want to configure the Enrollee, for example if the Enrollee wishes to be enrolled as an AP and
there are already enough APs in the network, the Configurator shall respond with a DPP Configuration Response indicating failure by adding the DPP Status field set to STATUS_CONFIGURE_FAILURE and wrapped data consisting of
the Enrollee’s nonce wrapped in ke"

### DIFF
--- a/inc/ec_ctrl_configurator.h
+++ b/inc/ec_ctrl_configurator.h
@@ -7,6 +7,11 @@
 // forward decl
 struct cJSON;
 
+typedef enum {
+    dpp_config_obj_bsta,
+    dpp_config_obj_ieee1905,
+} dpp_config_obj_type_e;
+
 class ec_ctrl_configurator_t : public ec_configurator_t {
 public:
     ec_ctrl_configurator_t(std::string mac_addr, send_chirp_func send_chirp_notification, send_encap_dpp_func send_prox_encap_dpp_msg,
@@ -92,7 +97,20 @@ private:
     std::pair<uint8_t*, size_t> create_recfg_auth_request();
     std::pair<uint8_t*, size_t> create_auth_confirm(std::string enrollee_mac, ec_status_code_t dpp_status, uint8_t* i_auth_tag);
     std::pair<uint8_t*, size_t> create_recfg_auth_confirm(std::string enrollee_mac, ec_status_code_t dpp_status);
-    std::pair<uint8_t*, size_t> create_config_response();
+    std::pair<uint8_t*, size_t> create_config_response(uint8_t dest_mac[ETH_ALEN], const uint8_t dialog_token, ec_status_code_t dpp_status);
+
+    /**
+     * @brief "Completes" a base DPP Configuration object
+     * 
+     * A base configuration object has all mandatory base fields filled out, including keys
+     * "wi-fi_tech", "discovery", and "credential"
+     * 
+     * @param base The base JSON object.
+     * @param conn_ctx The EC connection context.
+     * @param config_obj_type The type of DPP Configuration object we're filling out
+     * @return cJSON* DPP Configuration object on success, nullptr otherwise.
+     */
+    cJSON *complete_config_obj(cJSON *base, ec_connection_context_t& conn_ctx, dpp_config_obj_type_e config_obj_type);
 
     /**
      * @brief Maps Enrollee MAC (as string) to onboarded status. True if onboarded (now a Proxy Agent), false if still onboarding / onboarding failed.
--- a/inc/ec_ctrl_configurator.h
+++ b/inc/ec_ctrl_configurator.h
@@ -97,20 +97,29 @@ private:
     std::pair<uint8_t*, size_t> create_recfg_auth_request();
     std::pair<uint8_t*, size_t> create_auth_confirm(std::string enrollee_mac, ec_status_code_t dpp_status, uint8_t* i_auth_tag);
     std::pair<uint8_t*, size_t> create_recfg_auth_confirm(std::string enrollee_mac, ec_status_code_t dpp_status);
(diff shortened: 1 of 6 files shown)


inc/ec_ctrl_configurator.h
```diff
@@ -92,7 +97,20 @@ class ec_ctrl_configurator_t : public ec_configurator_t {
...
+     * 
+     * A base configuration object has all mandatory base fields filled out, including keys
+     * "wi-fi_tech", "discovery", and "credential"
+     * 
+     * @param base The base JSON object.
+     * @param conn_ctx The EC connection context.
+     * @param config_obj_type The type of DPP Configuration object we're filling out
+     * @return cJSON* DPP Configuration object on success, nullptr otherwise.
+     */
+    cJSON *complete_config_obj(cJSON *base, ec_connection_context_t& conn_ctx, dpp_config_obj_type_e config_obj_type);
```
bcarlson-dev, line 113: Maybe `finalize` instead of `complete`?

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -1116,7 +937,187 @@ std::pair<uint8_t *, size_t> ec_ctrl_configurator_t::create_recfg_auth_confirm(s
...
+        wrapped_attribs = ec_util::add_attrib(wrapped_attribs, &wrapped_len, ec_attrib_id_dpp_config_obj, bsta_config_obj_str);
+        return std::make_pair(wrapped_attribs, wrapped_len);
+    });
+
+    response_frame = reinterpret_cast<ec_gas_initial_response_frame_t*>(ec_util::copy_attrs_to_frame(reinterpret_cast<uint8_t*>(response_frame), sizeof(ec_gas_initial_response_frame_t), attribs, attribs_len));
+    free(attribs);
+    ASSERT_NOT_NULL(response_frame, {}, "%s:%d: Failed to copy attributes to DPP Configuration frame!\n", __func__, __LINE__);
+    response_frame->resp_len = static_cast<uint16_t>(attribs_len);
+    auto [encap_response_frame, encap_response_frame_len] = ec_util::create_encap_dpp_tlv(true, dest_mac, ec_frame_type_easymesh, reinterpret_cast<uint8_t*>(response_frame), frame_len + attribs_len);
+    ASSERT_NOT_NULL(encap_response_frame, {}, "%s:%d: Failed to alloc DPP Configuration frame!\n", __func__, __LINE__);
```
bcarlson-dev, line 1119: While it doesn't make it *as* clean, can you move the `create_encap_dpp_tlv` call outside of this function and into the caller in order to make this function a bit more clear that its generating the "response frame".

Similarly, could you rename the method to `create_config_response_frame`

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -1116,7 +937,187 @@ std::pair<uint8_t *, size_t> ec_ctrl_configurator_t::create_recfg_auth_confirm(s
...
 
-std::pair<uint8_t *, size_t> ec_ctrl_configurator_t::create_config_response()
+cJSON *ec_ctrl_configurator_t::complete_config_obj(cJSON *base, ec_connection_context_t& conn_ctx, dpp_config_obj_type_e config_obj_type)
 {
-    return {};
+    if (base == nullptr) {
+        printf("%s:%d: Based a nullptr base Configuration object\n", __func__, __LINE__);
+        return nullptr;
+    }
+    switch(config_obj_type) {
```
bcarlson-dev, line 946: Since all of the fields are the same except for `groups`, these field creations could be moved outside of the switch statement.

It would change things substantially (due to a need for a check on if the type is valid before hand) but it could avoid people reading in the future trying to play a game of spot the difference. 

You could put a switch at the top to the filtering if chosen:

```suggestion
    std::vector<std::unordered_map<std::string, std::string>> groups;
    switch(config_obj_type) {
        case dpp_config_obj_type_e::dpp_config_obj_bsta: {
           groups = {
                {{"groupID", "mapNW"}, {"netRole", "mapBackhaulSta"}}
            };
            break;
        }
        case dpp_config_obj_type_e::dpp_config_obj_ieee1905: {
           groups = {
                {{"groupID", "mapNW"}, {"netRole", "mapAgent"}}
            };
            break;
        }
        default:
            // Error
            return NULL;
    }

```

inc/em_agent.h
```diff
@@ -92,6 +92,8 @@ class em_agent_t : public em_mgr_t {
      */
     bool send_action_frame(uint8_t dest_mac[ETH_ALEN], uint8_t *action_frame, size_t action_frame_len, unsigned int frequency=0, unsigned int wait_time_ms=0) override;
 
+    bool can_onboard_additions_aps();
```
bcarlson-dev, line 95: Can you add a small docstring + typo
```suggestion
    bool can_onboard_additional_aps();
```

bcarlson-dev (changes requested):

inc/ec_ctrl_configurator.h
tuckerpo-cl: Certainly better than `complete`!

bcarlson-dev (approved):

PR #273, opened 2025-04-05
create_ieee1905_response_obj: include "cred" object

tuckerpo-cl:
`"cred"` object was created and populated but not added to the config object prior to returning -- fixes that

### DIFF
--- a/src/em/prov/em_provisioning.cpp
+++ b/src/em/prov/em_provisioning.cpp
@@ -818,6 +818,8 @@ cJSON *em_provisioning_t::create_ieee1905_response_obj(ec_connection_context_t *
         return nullptr;
     }
 
+    cJSON_AddItemToObject(dpp_configuration_object, "cred", credential_object);
+
     return dpp_configuration_object;
 }
 


src/em/prov/em_provisioning.cpp
```diff
@@ -818,6 +818,8 @@ cJSON *em_provisioning_t::create_ieee1905_response_obj(ec_connection_context_t *
         return nullptr;
     }
 
+    cJSON_AddItemToObject(dpp_configuration_object, "cred", credential_object);
```
bcarlson-dev, line 821: Does "credential_object" need to be deleted? Genuine question

bcarlson-dev (approved):

src/em/prov/em_provisioning.cpp
tuckerpo-cl: I'm 95% sure that `Add<x>ToObject` moves ownership to the object that's being added to, so no

bcarlson-dev: I thought so as well, I just wanted to make sure

PR #275, opened 2025-04-08
Fixed memory couruption caused by not allocating `okm` before passing into `hkdf`

bcarlson-dev:


tuckerpo-cl:
Wow, so `hkdf` wrote to a dangling pointer? Ouch!

tuckerpo-cl (approved):

PR #284, opened 2025-04-09
DPP EasyConnect: Implement required endianness conversions

apahl-cl:
The EasyConnect and EasyMesh specs require network byte order in some places and little endian in other places. This PR adds a bunch of byte order conversions throughout the EasyConnect code to improve spec conformity. 

One area that required a significant refactor was DPP attributes: ID and length fields are expected to be little endian. I introduced a new type, `ec_host_attribute_t`, to describe a DPP attribute where these fields are in host byte order. This will (hopefully) eliminate ambiguity about the byte ordering of a given instance. 

This code depends on the presence of `__BYTE_ORDER` and `__LITTLE_ENDIAN` as predefined macros, and it requires being able to import `byteswap.h`. All of these conditions are satisfied by glibc.

### DIFF
--- a/inc/ec_base.h
+++ b/inc/ec_base.h
@@ -229,7 +229,7 @@ typedef struct {
 } ec_credential_object_t;
 
 /**
- * @brief A DPP attribute as defined in EC 8.1. Can be sent/received over the network.
+ * @brief A DPP attribute as defined in EasyConnect 8.1. Can be sent/received over the network.
  * 
  * @paragraph DPP attributes' ID and length fields are required to be little 
  *            endian when transferred over the network. In general, data 
@@ -245,11 +245,11 @@ typedef struct {
  */
 typedef struct {
     /**
-     * @brief Identifies the type of the DPP attribute. Assumed to be little endian, as described in EC 8.1.
+     * @brief Identifies the type of the DPP attribute. Assumed to be little endian, as described in EasyConnect 8.1.
      */
     uint16_t attr_id;
     /**
-     * @brief Length of the following fields in the attribute. Assumed to be little endian, as described in EC 8.1.
+     * @brief Length of the following fields in the attribute. Assumed to be little endian, as described in EasyConnect 8.1.
      */
     uint16_t length;
     /**
@@ -288,7 +288,7 @@ typedef struct {
      */
     uint16_t length;
     /**
-     * @brief Points to the `ec_attribute_t` instance this `ec_attribute_t` instance was derived from.
+     * @brief Points to the `ec_net_attribute_t` instance this `ec_net_attribute_t` instance was derived from.
      */
     ec_net_attribute_t *original;
     /**
@@ -400,7 +400,7 @@ typedef struct {
  */
 #define ASSERT_OPT_HAS_VALUE_FREE3(x, ret, ptr1, ptr2, ptr3, errMsg, ...) \
     do { \
-        if(x.has_value()) { \
+        if(!x.has_value()) { \
             fprintf(stderr, errMsg, ## __VA_ARGS__); \
             void *_tmp1 = (ptr1); \
             void *_tmp2 = (ptr2); \
@@ -448,7 +448,7 @@ typedef struct {
  * @param errMsg Format string for error message
  * @param ... Additional arguments for the format string
  */
-#define ASSERT_OPT_HAS_VALUE(x, ret, errMsg, ...) ASSERT_MSG_FALSE(x.has_value(), ret, errMsg, ## __VA_ARGS__)
+#define ASSERT_OPT_HAS_VALUE(x, ret, errMsg, ...) ASSERT_MSG_TRUE(x.has_value(), ret, errMsg, ## __VA_ARGS__)
 
 #ifndef SSL_KEY
 #if OPENSSL_VERSION_NUMBER < 0x30000000L
--- a/inc/ec_base.h
+++ b/inc/ec_base.h
@@ -228,11 +228,74 @@ typedef struct {
       } creds;
 } ec_credential_object_t;
 
+/**
+ * @brief A DPP attribute as defined in EC 8.1. Can be sent/received over the network.
+ * 
+ * @paragraph DPP attributes' ID and length fields are required to be little 
+ *            endian when transferred over the network. In general, data 
+ *            transmitted over the network is big endian ("network byte 
+ *            order"), so DPP attributes must be treated as a different case.
+ *            Users of this struct MUST maintain the invariant that the 
+ *            `attr_id` and `length` are little endian. If anything needs to be
+ *            done with these values on the host, use an `ec_attribute_t`
+ *            instead.
+ * 
+ * @warning When doing any network operations involving DPP attributes, use `ec_net_attribute_t`, not `ec_attribute_t`. 
+ * @warning When executing logic on the host related to `attr_id` and `length`, use `ec_attribute_t`, not `ec_net_attribute_t`. 
+ */
 typedef struct {
+    /**
+     * @brief Identifies the type of the DPP attribute. Assumed to be little endian, as described in EC 8.1.
+     */
     uint16_t attr_id;
+    /**
+     * @brief Length of the following fields in the attribute. Assumed to be little endian, as described in EC 8.1.
+     */
     uint16_t length;
+    /**
+     * @brief Attribute-specific information fields. Endianness varies according to the specific DPP attribute type.
+     */
     uint8_t data[0];
-}__attribute__((packed)) ec_attribute_t;
+}__attribute__((packed)) ec_net_attribute_t;
+
+/**
+ * @brief Represents a DPP attribute that has been converted to host byte ordering. Not intended to be sent over the network. 
+ * 
+ * @paragraph This struct is intended to be used in any operations on DPP
+ *            attributes that occur entirely on the host. For example, after an
+ *            `ec_net_attribute_t` is read from a frame received from
+ *            the network, it must be converted to an `ec_attribute_t` to
+ *            ensure that the `attr_id` and `length` use host byte ordering. 
+ * 
+ * @paragraph Note that this struct includes a pointer, 
+ *            `ec_net_attribute_t *original`. This MUST ALWAYS point to the
+ *            `ec_net_attribute_t` instance that was used to derive this
+ *            `ec_attribute_t` instance. `original` is used in pointer
+ *            arithmetic based on its position in a frame. 
+ * 
+ * @note `ec_net_attribute_t *original` points to the `ec_attribute_t` instance this was derived from.
+ * 
+ * @warning When doing any network operations involving DPP attributes, use `ec_net_attribute_t`, not `ec_attribute_t`. 
+ * @warning When executing logic on the host related to `attr_id` and `length`, use `ec_attribute_t`, not `ec_net_attribute_t`. 
+ */
+typedef struct {
+    /**
+     * @brief Identifies the type of the DPP attribute. Assumed to be stored with host byte ordering.
+     */
+    uint16_t attr_id;
+    /**
+     * @brief Length of the following fields in the attribute. Assumed to be stored with host byte ordering.
+     */
+    uint16_t length;
+    /**
+     * @brief Points to the `ec_attribute_t` instance this `ec_attribute_t` instance was derived from.
+     */
+    ec_net_attribute_t *original;
+    /**
+     * @brief Shorthand for `this.original->data`. Must always equate to `this.original->data`. 
+     */
+    uint8_t *data;
+} ec_attribute_t;
 
 typedef struct {
     uint8_t category;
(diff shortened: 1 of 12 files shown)


apahl-cl:
@bcarlson-dev @tuckerpo-cl

inc/ec_base.h
```diff
@@ -325,6 +354,68 @@ typedef struct {
...
+ * @brief Asserts that a std::optional has a value, and if it doesn't, frees up to 3 pointers and returns a value
+ * @param x The std::optional to check for a value
+ * @param ret The value to return if x is nullopt
+ * @param ptr1 First pointer to free (can be NULL)
+ * @param ptr2 Second pointer to free (can be NULL) 
+ * @param ptr3 Third pointer to free (can be NULL)
+ * @param errMsg Format string for error message
+ * @param ... Additional arguments for the format string
+ */
+#define ASSERT_OPT_HAS_VALUE_FREE3(x, ret, ptr1, ptr2, ptr3, errMsg, ...) \
```
bcarlson-dev, line 367: I don't think these are needed. If you replace the `if(x == NULL)` with `if(!x)` in the original one, it should still do the same function

tests/test_util.cpp
```diff
@@ -0,0 +1,66 @@
+#include <gtest/gtest.h>
+#include <string>
+
+#include "util.h" 
+#include <netinet/in.h>
+
+TEST(EmUtilTest, TestStringSplit) {
```
bcarlson-dev, line 7: Can you add a few comments to these tests to clarify for those reading it

inc/ec_base.h
bcarlson-dev: If you want the names for clarity, then it might be better to make that change and have them just be defined based on the `ASSERT_NOT_NULL_FREE*` macros

inc/util.h
```diff
@@ -138,6 +152,26 @@ std::pair<uint8_t, uint8_t> em_freq_to_chan(unsigned int frequency, const std::s
...
 
+/**
+ * @brief Retrieve a network byte ordered uint16_t from an address and convert to host byte ordering
+ * 
+ * @param ptr Pointer to an address containing an unsigned 16 bit integer
+ * @return uint16_t The integer at *ptr, converted to host byte ordering, or 0 if ptr is NULL.
+ * 
+ * @note `ptr` is `void*` because this is typically invoked on data that must be reinterpreted from its assigned type.
+ */
+uint16_t deref_net_uint16_to_host(const void* const ptr);
```
bcarlson-dev, line 163: Can you make reference to `htons` and `ntohs` in the documentation for these functions just so people can easily look and be like "okay that is what it is doing"

bcarlson-dev (changes requested):

apahl-cl:
Pushed a bunch of changes based on your feedback @bcarlson-dev:
- Renamed `ec_attribute_t` to `ec_net_attribute_t`
- Renamed `ec_host_attribute_t` to `ec_attribute_t`
- Added detailed Doxygen comments for each `*_attribute_t`
- Replaced a bunch of `std::optional<...>` with `auto`
- Fixed the use of`free(unwrapped_attrs)` in a few places
- Added a `data` element to the new `ec_attribute_t` as a shorthand for `ec_attribute_t.original->data`
- Added unit testing for `ec_attribute_t.original`
- Added explanatory comments to the `EmUtilByteOrderTest` unit tests

inc/util.h
apahl-cl: Let me know if you had more in mind than what I added here. I basically just wrote enough to be coherent and greppable.

apahl-cl:
I made the `ec_attribute_t.original` unit testing a separate commit so it's a bit easier to see in isolation.

inc/ec_base.h
```diff
@@ -228,11 +228,74 @@ typedef struct {
       } creds;
 } ec_credential_object_t;
 
+/**
+ * @brief A DPP attribute as defined in EC 8.1. Can be sent/received over the network.
```
bcarlson-dev, line 232: In header files when referencing specifications, can you spell out the spec (EasyConnect) instead of abreviating (EC)
```suggestion
 * @brief A DPP attribute as defined in EasyConnect 8.1. Can be sent/received over the network.
```

inc/ec_base.h
```diff
@@ -228,11 +228,74 @@ typedef struct {
...
+    /**
+     * @brief Identifies the type of the DPP attribute. Assumed to be stored with host byte ordering.
+     */
+    uint16_t attr_id;
+    /**
+     * @brief Length of the following fields in the attribute. Assumed to be stored with host byte ordering.
+     */
+    uint16_t length;
+    /**
+     * @brief Points to the `ec_attribute_t` instance this `ec_attribute_t` instance was derived from.
```
bcarlson-dev, line 291: Missed in the renaming
```suggestion
     * @brief Points to the `ec_net_attribute_t` instance this `ec_attribute_t` instance was derived from.
```

inc/ec_util.h
```diff
@@ -519,20 +518,20 @@ class ec_util {
...
+	 *         which can then be fetched via `get_attrib`, along with the length of that buffer.
 	 *         The buffer is NULL and the size is 0 on failure.
 	 *
 	 * @warning The caller is responsible for freeing the memory returned by this function.
 	 *
-	 * @note Forwards to `unwrap_wrapped_attrib(ec_attribute_t *wrapped_attrib, uint8_t *frame, size_t frame_len, uint8_t *frame_attribs, bool uses_aad, uint8_t *key);`
+	 * @note Forwards to `unwrap_wrapped_attrib(ec_attribute_t wrapped_attrib, uint8_t *frame, size_t frame_len, uint8_t *frame_attribs, bool uses_aad, uint8_t *key);`
 	 */
-	static std::pair<uint8_t*, uint16_t> unwrap_wrapped_attrib(ec_attribute_t* wrapped_attrib, ec_frame_t *frame, bool uses_aad, uint8_t* key);
+	static std::pair<uint8_t*, uint16_t> unwrap_wrapped_attrib(ec_attribute_t wrapped_attrib, ec_frame_t *frame, bool uses_aad, uint8_t* key);
```
bcarlson-dev, line 534: Performance improvement, can you pass by `const` reference here since we are no longer using pointers.
```suggestion
	static std::pair<uint8_t*, uint16_t> unwrap_wrapped_attrib(const ec_attribute_t& wrapped_attrib, ec_frame_t *frame, bool uses_aad, uint8_t* key);
```

inc/ec_util.h
```diff
@@ -551,7 +550,7 @@ class ec_util {
 	 *
 	 * @note Ensure that the key is valid and the frame is correctly set up before calling this function.
 	 */
-	static std::pair<uint8_t*, uint16_t> unwrap_wrapped_attrib(ec_attribute_t *wrapped_attrib, uint8_t *frame, size_t frame_len, uint8_t *frame_attribs, bool uses_aad, uint8_t *key);
+	static std::pair<uint8_t*, uint16_t> unwrap_wrapped_attrib(ec_attribute_t wrapped_attrib, uint8_t *frame, size_t frame_len, uint8_t *frame_attribs, bool uses_aad, uint8_t *key);
```
bcarlson-dev, line 553: Same `const` ref comment here

src/em/prov/easyconnect/ec_util.cpp
```diff
@@ -24,25 +24,35 @@ void ec_util::init_frame(ec_frame_t *frame)
     frame->crypto_suite = 0x01; // Section 3.3 (Currently only 0x01 is defined)
 }
 
-ec_attribute_t *ec_util::get_attrib(uint8_t *buff, size_t len, ec_attrib_id_t id)
+std::optional<ec_attribute_t> ec_util::get_attrib(uint8_t *buff, size_t len, ec_attrib_id_t id)
```
bcarlson-dev, line 27: Based on the prior comment, saying that to pass in `ec_attribute_t` as a `const` ref, it might make it easier to make the return type here be a `const std::optional<ec_attribute_t>`. It would also make it more clear since none of the `ec_attribute_t`s parameters should be modified since it is just a helper struct. The actual buffer data is in `ec_net_attribute_t` and (therefore) `data` (whose pointer also can't be modified with a `const`).

tests/test_ec_attrib.cpp
```diff
@@ -105,30 +118,33 @@ TEST_F(ECUtilAttributeTest, GetAttribute) {
...
-    ec_attribute_t* attr2 = ec_util::get_attrib(buffer, buffer_len, ec_attrib_id_init_bootstrap_key_hash);
-    ASSERT_NE(attr2, nullptr);
+    auto attr2 = ec_util::get_attrib(buffer, buffer_len, ec_attrib_id_init_bootstrap_key_hash);
+    ASSERT_NE(attr2, std::nullopt);
     EXPECT_EQ(attr2->attr_id, ec_attrib_id_init_bootstrap_key_hash);
     EXPECT_EQ(attr2->length, sizeof(uint16_t));
+    EXPECT_EQ(attr2->original, reinterpret_cast<ec_net_attribute_t*>(test_attrib_2));
 
     uint16_t data = 0;
     memcpy(&data, attr2->data, sizeof(uint16_t));
```
bcarlson-dev, line 135: We can use one of your new `deref` calls if you want here unless the goal is to keep those functions out intentionally

tests/test_ec_attrib.cpp
```diff
@@ -45,22 +45,35 @@ class ECUtilAttributeTest : public ::testing::Test {
...
     ec_frame_t* frame = nullptr;
     uint8_t* buffer = nullptr;
 
+    const uint8_t test_val_1 = 0x42;
+    const uint16_t test_val_2 = 0x1234;
+    const std::string test_val_3 = "test_string";
+
+    uint8_t *test_attrib_1;
+    uint8_t *test_attrib_2;
+    uint8_t *test_attrib_3;
```
bcarlson-dev, line 75: Might be worth while to just set these to `NULL` in the `TearDown` function

bcarlson-dev (changes requested):
Honestly it looks pretty near perfect. Since this is a pretty crucial piece of code I want to be really through so I appreciate your patience.

tests/test_ec_attrib.cpp
apahl-cl: I don't think that will work here actually - the new `deref` calls use `ntohs`, but the input is network byte ordered and we want the output to keep that.

src/em/prov/easyconnect/ec_util.cpp
apahl-cl: I tested returning `const std::optional<ec_attribute_t>` from `get_attrib`. It still allows modifications - I think the intuition is that the subsequent assignment, `auto foo = get_attrib(...)`, counts as a copy and discards the const qualifier (not 100% sure though).

Protecting the struct from accidental modifications seems a bit complicated - I think what we really want is to return `std::optional<const ec_attribute_t>`. Or, for completeness, we can return `const std::optional<const ec_attribute_t>` AND also use `const auto foo = ...` on the caller's side.

apahl-cl: Okay, I've changed the return value to `const std::optional<const ec_attribute_t>` and I'm using `const auto foo = ...` on the caller side. We can adjust further if needed.

apahl-cl:
Implemented most of the changes from Ben's latest code review. Modified `TEST_F(ECUtilAttributeTest, NullBuffer)` (`test_ec_attrib.cpp:261`) because an attribute was being reassigned in that test.

src/em/prov/easyconnect/ec_util.cpp
```diff
@@ -24,25 +24,35 @@ void ec_util::init_frame(ec_frame_t *frame)
     frame->crypto_suite = 0x01; // Section 3.3 (Currently only 0x01 is defined)
 }
 
-ec_attribute_t *ec_util::get_attrib(uint8_t *buff, size_t len, ec_attrib_id_t id)
+const std::optional<const ec_attribute_t> ec_util::get_attrib(uint8_t *buff, size_t len, ec_attrib_id_t id)
```
bcarlson-dev, line 27: So I think you were right and we don't need the `const` for the optional return type. Removing it will also remove the need for the `const auto` as well. If I am not mistaken, keeping the `const ec_attribute_t` should make it so that when you do the `*some_attrib` later on you won't have to do a cast so we should be good if you remove this and switch back from the `const auto` to the `auto`?
```suggestion
std::optional<const ec_attribute_t> ec_util::get_attrib(uint8_t *buff, size_t len, ec_attrib_id_t id)
```

bcarlson-dev (approved):

tuckerpo-cl:
Thanks for plugging all the leaks I created 🙏  C moment!

apahl-cl:
> Thanks for plugging all the leaks I created 🙏 C moment!

Glad i could help a bit, and thanks to Ben too for plugging all the leaks I created to get this far!

PR #285, opened 2025-04-09
Fix warning for controller

prashant-singh1:
Reason for change: 1) Handled static and const cast warnings 
2) Handled few unused variable warnings
3) Handled remaining missing default case for switch statement.
Test Procedure: Checked and compared the warning before and after fix, observed around 255 warning is fixed. Also compared the run time logs for OneWifi, controller and agent with and without changes and did not observe any additional logs/error. Tested the changes by changing the SSID and radio channel through CLI observed its works fine. 
Risks: Medium
Priority: P1

### DIFF
--- a/src/dm/dm_network_ssid_list.cpp
+++ b/src/dm/dm_network_ssid_list.cpp
@@ -199,7 +199,7 @@ void dm_network_ssid_list_t::delete_list()
 
 bool dm_network_ssid_list_t::operator == (const db_easy_mesh_t& obj)
 {
-	dm_network_ssid_t *pnet_ssid = (dm_network_ssid_t*)&obj;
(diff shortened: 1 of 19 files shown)


src/dm/dm_network_ssid_list.cpp
```diff
@@ -199,7 +199,7 @@ void dm_network_ssid_list_t::delete_list()
 
 bool dm_network_ssid_list_t::operator == (const db_easy_mesh_t& obj)
 {
-	dm_network_ssid_t *pnet_ssid = (dm_network_ssid_t*)&obj;
+	dm_network_ssid_t *pnet_ssid = const_cast<dm_network_ssid_t*>(reinterpret_cast<const dm_network_ssid_t*>(&obj));
```
amarnathhullur, line 202: what is the need for const_cast again?

amarnathhullur (approved):

PR #294, opened 2025-04-09
Fixed the PEM file not reading in the private and public keys correctly.

bcarlson-dev:


src/em/crypto/em_crypto.cpp
```diff
@@ -1237,23 +1239,13 @@ SSL_KEY *em_crypto_t::read_keypair_from_pem(const std::string &file_path) {
     fp = fopen(file_path.c_str(), "rb");
     ASSERT_NOT_NULL(fp, NULL, "%s:%d Failed to open file (%s)", __func__, __LINE__, file_path.c_str());
     
-    // Read private key first
+    // Read private key - this will contain both private and public key information
     pkey = PEM_read_PrivateKey(fp, NULL, NULL, NULL);
     if (!pkey) {
         printf("%s:%d Failed to read private key from PEM file\n", __func__, __LINE__);
         goto err;
     }
```
tuckerpo-cl, line 1247: Wouldn't this only read until `-----END PRIVATE KEY-----`? Do we also need `PEM_read_PUBKEY`?

tuckerpo-cl (changes requested):

src/em/crypto/em_crypto.cpp
bcarlson-dev: That's what I thought as well but each of these functions write over the pkey completely and when you read the private key it actually handles the public key details for you (since those are computed from the private key after all)

tuckerpo-cl (approved):

PR #295, opened 2025-04-10
Fixed bugs and added tests for AES SIV and EC Attrib

bcarlson-dev:


mhughesacn:
Hi @bcarlson-dev : Please confirm that the pem file which you removed contains only test data.
Martin H, RDK CMF Compliance Team

bcarlson-dev:
@mhughesacn 
Yes it is only test data. I generated it for the first time through the "openssl" CLI tool and it will only be called through tests.

PR #299, opened 2025-04-10
Fixed the ethernet backhaul ethernet issue and handling parallel renew for bands

AswiniViswanathan:
Fixed the ethernet backhaul ethernet issue and handling parallel renew for bands

    Tesing:
    1)ethernet backhaul : private and IOT change for 10 iteration
    2)Wifi backhaul: Cannot test the 2.5 and 5G simultaneously for any other haultype except private .
    We discussed and agreed we should not have to change the backhaul AL mac interface which in our case its private.
    We cannot have the remote agent also with IOT of 2.4 and 5G simultaneously .
    So autoconfig renew for both 2.5 and 5G cannot be test with wifi backhaul in raspi setup


src/orch/em_orch.cpp
```diff
@@ -243,16 +243,77 @@ bool em_orch_t::eligible_for_active(em_cmd_t *pcmd)
...
     em_short_string_t key;
     em_cmd_type_t	type;
 
-    type = em_cmd_t::bus_2_cmd_type(etype);	
+    type = em_cmd_t::bus_2_cmd_type(evt->type);
 
     snprintf(key, sizeof(em_short_string_t), "%d", type);
 
+    if (type == em_cmd_type_cfg_renew ) {
+        is_cmd_type_renew_in_progress(evt);
```
amarnathhullur, line 315: should have return here, no need of proceeding further

src/orch/em_orch.cpp
```diff
@@ -243,16 +243,77 @@ bool em_orch_t::eligible_for_active(em_cmd_t *pcmd)
...
+	if ((stats = (em_cmd_stats_t *)hash_map_get(m_cmd_map, key)) != NULL) {
+	   printf("%s:%d: Command of type: %d actively executing\n", __func__, __LINE__, type);
+		// Go through the pending queue and check if radio mac match
+		for (i = queue_count(m_pending) - 1; i >= 0; i--) {
+			pcmd = (em_cmd_t *)queue_peek(m_pending, i);
+			if (pcmd->m_type == type) {
+				for (j = queue_count(pcmd->m_em_candidates) - 1; j >= 0; j--) {
+					em = (em_t *)queue_peek(pcmd->m_em_candidates, j);
+					dm_easy_mesh_t::macbytes_to_string(em->get_radio_interface_mac(), mac_str);
+					printf("%s:%d: ASH %s radio config renew\n", __func__, __LINE__, mac_str);
```
amarnathhullur, line 274: Remove ASH in the printf

src/orch/em_orch.cpp
```diff
@@ -243,16 +243,77 @@ bool em_orch_t::eligible_for_active(em_cmd_t *pcmd)
...
+			}
+		}
+		// go though active queue and check if radio mac match
+		for (i = queue_count(m_active) - 1; i >= 0; i--) {
+			pcmd = (em_cmd_t *)queue_peek(m_active, i);
+			if (pcmd->m_type == type) {
+				for (j = queue_count(pcmd->m_em_candidates) - 1; j >= 0; j--) {
+					em = (em_t *)queue_peek(pcmd->m_em_candidates, j);
+					dm_easy_mesh_t::macbytes_to_string(em->get_radio_interface_mac(), mac_str);
+					printf("%s:%d: ASH %s radio config renew\n", __func__, __LINE__, mac_str);
```
amarnathhullur, line 290: Same here

src/orch/em_orch.cpp
AswiniViswanathan: addressed

src/orch/em_orch.cpp
AswiniViswanathan: Addressed

src/orch/em_orch.cpp
AswiniViswanathan: Addressed

amarnathhullur (approved):

PR #301, opened 2025-04-10
Co-located agent awareness and wrong EC frame type 

bcarlson-dev:
Fixed a colocated agent not knowing that it is co-located and fixed authentication responses being created with the wrong type.

### DIFF
--- a/src/agent/em_agent.cpp
+++ b/src/agent/em_agent.cpp
@@ -462,7 +462,12 @@ void em_agent_t::handle_recv_wfa_action_frame(em_bus_event_t *evt)
             em_printfout("Dest radio node MAC '" MACSTRFMT "', al_node radio MAC '" MACSTRFMT"'\n", MAC2STR(dest_radio_node->get_radio_interface_mac()), MAC2STR(al_node->get_radio_interface_mac()));
             dest_al_same = (memcmp(dest_radio_node->get_radio_interface_mac(), al_node->get_radio_interface_mac(), ETH_ALEN) == 0);
         }
-        em_printfout("Dest MAC '%s', dest_al_same=%d, is_bcast=%d\n", dest_mac_str, dest_al_same, is_bcast);
+
+        auto ctrl_al = m_data_model.get_controller_interface_mac();
+        auto agent_al = m_data_model.get_agent_al_interface_mac();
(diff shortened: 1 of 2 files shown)


src/agent/em_agent.cpp
```diff
@@ -474,20 +479,17 @@ void em_agent_t::handle_recv_wfa_action_frame(em_bus_event_t *evt)
...
         }
+        em_printfout("Did not find an EM node for action frame!");
     }
     default:
         break;
     }
-
-    if (!found_em) {
-        em_printfout("Did not find an EM node for action frame!\n");
-    }
```
tuckerpo-cl, line 490: `bool found_em = false;` at 455 can be deleted

src/agent/em_agent.cpp
```diff
@@ -462,7 +462,12 @@ void em_agent_t::handle_recv_wfa_action_frame(em_bus_event_t *evt)
             em_printfout("Dest radio node MAC '" MACSTRFMT "', al_node radio MAC '" MACSTRFMT"'\n", MAC2STR(dest_radio_node->get_radio_interface_mac()), MAC2STR(al_node->get_radio_interface_mac()));
             dest_al_same = (memcmp(dest_radio_node->get_radio_interface_mac(), al_node->get_radio_interface_mac(), ETH_ALEN) == 0);
         }
-        em_printfout("Dest MAC '%s', dest_al_same=%d, is_bcast=%d\n", dest_mac_str, dest_al_same, is_bcast);
+
+        auto ctrl_al = m_data_model.get_controller_interface_mac();
+        auto agent_al = m_data_model.get_agent_al_interface_mac();
+        bool is_colocated = (memcmp(ctrl_al, agent_al, ETH_ALEN) == 0);
+
+        em_printfout("Dest MAC '%s', dest_al_same=%d, is_bcast=%d, is_colocated=%d", dest_mac_str, dest_al_same, is_bcast, is_colocated);
```
tuckerpo-cl, line 470: Do we want to `root cause` fix this so that `m_data_model.get_colocated()` isn't just straight up lying? I'm OK with letting that be "out of scope" for us if not.

tuckerpo-cl (changes requested):

tuckerpo-cl (dismissed):

tuckerpo-cl (approved):

PR #314, opened 2025-04-15
Fixed addressing for rdkb-ieee1905.1-service-access with co-located agent+controller

bcarlson-dev:
- If a controller and an agent are on the same device (and will therefore be sharing an AL_MAC), they can now communicate with each other.
- Implemented a try-catch to prevent one bad or out of order packet from crashing the agent/controller

### DIFF
--- a/src/em/em_mgr.cpp
+++ b/src/em/em_mgr.cpp
@@ -373,9 +373,18 @@ void em_mgr_t::nodes_listener()
         while (em != NULL) {
             if (em->is_al_interface_em() == true) {
 #ifdef AL_SAP
-                AlServiceDataUnit sdu = g_sap->serviceAccessPointDataIndication();
-                std::vector<unsigned char> payload = sdu.getPayload();
-                proto_process(payload.data(), payload.size(), em);
+                try{
+                    AlServiceDataUnit sdu = g_sap->serviceAccessPointDataIndication();
+                    std::vector<unsigned char> payload = sdu.getPayload();
+                    proto_process(payload.data(), payload.size(), em);
+                } catch (const AlServiceException& e) {
+                    if (e.getPrimitiveError() == PrimitiveError::InvalidMessage) {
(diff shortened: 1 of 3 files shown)


src/em/em_mgr.cpp
```diff
@@ -373,9 +373,17 @@ void em_mgr_t::nodes_listener()
...
-                proto_process(payload.data(), payload.size(), em);
+                try{
+                    AlServiceDataUnit sdu = g_sap->serviceAccessPointDataIndication();
+                    std::vector<unsigned char> payload = sdu.getPayload();
+                    proto_process(payload.data(), payload.size(), em);
+                } catch (const AlServiceException& e) {
+                    if (e.getPrimitiveError() == PrimitiveError::IndicationFailed) {
+                        em_printfout("%s. Dropping packet", e.what());
+                    } else {
+                        throw e; // rethrow the exception if it's not an indication failure
```
tuckerpo-cl, line 384: Could we log `e.what()` prior to re-throwing, here? Since that'd lead to a crash (unless we start `try { } catch(...) {}`-ing everywhere which I'd really rather not do, might be useful to see what the actual exception was in stdout

tuckerpo-cl (changes requested):

src/em/em_mgr.cpp
tuckerpo-cl: Yeah definitely should kill agent/controller if the tunnel dies, but whether or not an uncaught exception's `.what()` gets printed is runtime implementation dependent, so if this is ever meant to run on some goofy hardware with a really slim libc implementation, it potentially _wouldn't_ log it (GCC/glibc and Clang both do, unsure of MSVC). Just thinking of portability

bcarlson-dev: That's totally fair. I'll make the change

tuckerpo-cl (approved):

PR #318, opened 2025-04-16
Fix warnings from controller side.

prashant-singh1:
Fix warning for controller (unified-wifi-mesh)

Reason for change: 1) Handled some datatype conversion related warnings 
2) Handled warnings related to makefile flags and separated C and CPP flags. 
Test Procedure: Checked and compared the warning before and after fix, observed around 94 warning is fixed from controller compilation. 
Also compared the run time logs for OneWifi, controller and agent with and without changes and did not observed any additional logs/error. 
Tested the changes by changing the SSID, Passphrase and radio channel through CLI observed its works fine. 
Risks: Medium
Priority: P1


src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -942,6 +942,7 @@ std::pair<uint8_t *, size_t> ec_ctrl_configurator_t::create_recfg_auth_confirm(s
     uint8_t trans_id = 0;
     ec_dpp_reconfig_flags_t reconfig_flags = {
         .connector_key = 1, // DONT REUSE
+        .reserved = 0,
```
amarnathhullur, line 945: Addition of this causes issue in openwrt based build. Please check to have an alternate solution

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -942,6 +942,7 @@ std::pair<uint8_t *, size_t> ec_ctrl_configurator_t::create_recfg_auth_confirm(s
...
     };
```
bcarlson-dev, line 946:
```suggestion
    ec_dpp_reconfig_flags_t reconfig_flags;
    memset(&reconfig_flags, 0, sizeof(ec_dpp_reconfig_flags_t));
    reconfig_flags.connector_key = 1;
```
@amarnathhullur  I believe this should resolve both the warning and the openwrt based build issue

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
prashant-singh1: @amarnathhullur @bcarlson-dev 
I modified and committed as below and seems build is also fine, let me know if me know if any further modification required.
Screenshot 2025-04-17 203840

bcarlson-dev:
I just fixed the deprecation warnings in the AES-SIV file and I was worried since I had to change so much that your fixes for those warnings might no longer cleanly apply so I fixed those as well. Can you revert the changes you made to the `aes_siv.c` file? Those warnings will be fixed with my PR

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
amarnathhullur: yes probably this should fix.

prashant-singh1:
Hi @bcarlson-dev
I have reverted changes aes_siv.c file as requested, also modified the changes for ec_ctrl_configurator.cpp as per your suggestion.

@amarnathhullur Request you to kindly review and approve the changes.

amarnathhullur (approved):

PR #327, opened 2025-04-21
Enable 4 address mode by default in EasymeshCfg.json file

amarnathhullur:
This is to enable 4 address station mode by default in EasymeshCfg.json. 
This parameter is read by rdk-wifi-hal to configure station in 4 address mode.


amarnathhullur:
This is associated with PR - https://github.com/rdkcentral/rdk-wifi-hal/pull/150

bcarlson-dev:
Given the discussion in last week's Wi-Fi WG meeting about WDS being a largely outdated term with "4 address mode" being the preferred term, should the configuration option be named something different? (Although I understand the comments specify it as 4 addr mode)

amarnathhullur:
> Given the discussion in last week's Wi-Fi WG meeting about WDS being a largely outdated term with "4 address mode" being the preferred term, should the configuration option be named something different? (Although I understand the comments specify it as 4 addr mode)

Can I change to 4addr_mode, do you have any other preference?

bcarlson-dev:
I think it could be best as a boolean `4addr_mode_enabled : true` which would avoid any confusion about their being an integer instead of a boolean. This would also remove the need for the equality check in `rdk-wifi-hal`

bcarlson-dev (approved):
Assuming the `Makefile.am` changes resolved the issue we discussed before then it LGTM

PR #329, opened 2025-04-21
Allows for a mesh backhaul STA to be associated to a BSS through UWM

bcarlson-dev:
- Processes the mesh_backhaul_sta subdocs received 
- Adds `mesh_sta_passphrase` to specify the passphrase for the BSS

### DIFF
--- a/inc/em_agent.h
+++ b/inc/em_agent.h
@@ -514,6 +514,17 @@ public:
 	 * @note Ensure that the event structure is properly initialized before passing it to this function.
 	 */
 	void handle_onewifi_private_cb(em_bus_event_t *evt);
+
+	/**!
+	 * @brief Handles the mesh station callback for OneWiFi events.
+	 *
+	 * This function processes the event received from the OneWiFi bus and performs necessary actions.
+	 *
+	 * @param[in] evt Pointer to the event structure containing OneWiFi event data.
+	 *
+	 * @note Ensure that the event structure is properly initialized before passing it to this function.
+	 */
+	void handle_onewifi_mesh_sta_cb(em_bus_event_t *evt);
 	
 	/**!
 	 * @brief Handles the OneWiFi radio callback event.
--- a/inc/em_base.h
+++ b/inc/em_base.h
@@ -2242,6 +2242,7 @@ typedef struct {
     bool    multi_bssid;
     bool    transmitted_bssid;
     em_eht_operations_bss_t eht_ops;
+    em_short_string_t mesh_sta_passphrase;
 
     // Extra vendor information elements for the BSS
     // @note Don't manually allocate, use the helper functions to add/remove elements 
--- a/inc/em_base.h
+++ b/inc/em_base.h
@@ -2579,6 +2579,7 @@ typedef enum {
     em_bus_event_type_m2_tx,
     em_bus_event_type_topo_sync,
     em_bus_event_type_onewifi_private_cb,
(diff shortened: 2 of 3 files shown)


AswiniViswanathan:
Backhaul STA support in webconfig is removed since, controller config should not impact the mesh backhaul sta config and contoller should not do any change for this vap . According to the discussions Backhaul Mesh STA config is updated only through the config file[/nvram/EasymeshCfg.json] and Interface.json file. Introducing mesh STA in agent and webconfig will impact the existing wifi backhaul on boarding.  To bring up meshsta with a bss vap we can declare it in Interface.json file of Onewifi

bcarlson-dev:
> Backhaul STA support in webconfig is removed since, controller config should not impact the mesh backhaul sta config and contoller should not do any change for this vap . According to the discussions Backhaul Mesh STA config is updated only through the config file[/nvram/EasymeshCfg.json] and Interface.json file. Introducing mesh STA in agent and webconfig will impact the existing wifi backhaul on boarding. To bring up meshsta with a bss vap we can declare it in Interface.json file of Onewifi

So a few things here. 
1. The controller is not necessarily the one changing anything. In **non-hardcoded** mesh on-boarding, the agent/device that is attempting to get on the network (that has never connected through EasyMesh, 1905, 802.11, 802.3, or any other mechanism to the controller) must find a way to know what network it should be connecting to. It finds this out through DPP. Once it knows the credentials (SSID, PSK, AKMs), then it knows what network to join and will try and join it. Until that point, there is no way for it to join it unless you hardcoded the data into a JSON file.
2. This needs to be done at any point (from OneWifi's perspective). It cannot be just based on the value that is present in the JSON file at OneWifi startup (although we can write to it as well in the case that the device reboots **after provisioning**). 
3. This does not impact the existing and current onboarding method. I have tested it. This code being present still lets OneWifi take the `BackhaulSSID` and password from the JSON (or the default ones) and associate to it without the `unified-wifi-mesh` **agent** even running.  Additionally, if the existing "onboarding" method is in-fact in-place, the `bssid_is_changed` event will not trigger because (as shown below), when a STA is already associated (the below output is via the existing method but on this branch), the BSSID, SSID, MAC, and Passphrase are the same that the agent would be applying via webconfig. 
4. Lastly, if the bSTA connection to the BSS **did** need to change (which is a logical event to occur for DPP Reconfiguration) then the existing state machine in OneWifi handles `connected` -> `connected` transitions. 

Overall, we need to have a runtime version to do this, this does not conflict with the existing method, and the existing method requires the credentials to have always been located in the JSON file since the start of OneWifi so there is no way to actually onboard without knowing them before hand. A change like will be necessary at some point.

```
			"VapName":	"mesh_sta_2g",
			"BridgeName":	"brlan0",
			"VapMode":	1,
			"RadioIndex":	0,
			"Exists":	true,
			"SSID":	"RPI_RDKB-AP0",
			"BSSID":	"1c:bf:ce:fc:21:76",
			"MAC":	"94:18:65:5c:e7:ac",
			"Enabled":	true,
			"ConnectStatus":	true,
			"Security":	{
				"Mode":	"WPA2-Personal",
				"MFPConfig":	"Disabled",
				"EncryptionMethod":	"AES",
				"Wpa3_transition_disable":	false,
				"Passphrase":	"12345678",
				"KeyId":	"",
				"RekeyInterval":	0
			},
			"ScanParameters":	{
				"Period":	10,
				"Channel":	0
			}
```

tuckerpo-cl (approved):

PR #330, opened 2025-04-22
Add subscription for association status

tuckerpo-cl:
If the Enrollee's association to the mesh bSTA fails and the Configurator has requested a Configuration Result Status Frame, we can asynchronously send this frame based on the result of Configuration which is provided by the OneWifi subscription.

Also corrects a previous misunderstanding from the EasyConnect spec --

Upon receiving the final Configuration Response frame, the Enrollee shall immediately send a DPP Configuration Result frame prior to any association attempt.

Depends on https://github.com/rdkcentral/OneWifi/pull/287 for the subscribable path

### DIFF
--- a/inc/ec_enrollee.h
+++ b/inc/ec_enrollee.h
@@ -175,6 +175,18 @@ public:
 	 */
 	bool add_presence_announcement_freq(unsigned int freq);
 
+	/**
+	 * @brief Handles an association status event.
+	 * 
+	 * If this event maps to our bSTA's association attempt, check the association status and send a Configuration Status Result frame
+	 * 
+	 * If this event does NOT map to our bSTAs association attempt, just throw it away
+	 * 
+	 * @param sta_data The STA data containing association status
+	 * @return true on success, otherwise false
+	 */
+	bool handle_assoc_status(const rdk_sta_data_t &sta_data);
+
 private:
 
     
@@ -420,6 +432,13 @@ private:
 	// Key -> sender's MAC + dialog_token (example: aabbcceeddff_42) as a string
 	// Value -> buffer for re-assembling GAS frame fragments
 	std::unordered_map<std::string, gas_fragment_buffer_t> m_gas_fragments;
+
+	/**
+	 * @brief Map of BSSIDs (as strings) that we are awaiting association status for
+	 * Key: BSSID (as string)
+	 * Value: The MAC of the node we're trying to associate to
+	 */
+	std::unordered_map<std::string, mac_addr_t> m_awaiting_assoc_status = {};
 };
 
 #endif // EC_ENROLLEE_H
--- a/src/em/prov/easyconnect/ec_enrollee.cpp
+++ b/src/em/prov/easyconnect/ec_enrollee.cpp
@@ -571,21 +571,15 @@ bool ec_enrollee_t::handle_config_response(uint8_t *buff, size_t len, uint8_t sa
         return false;
     }
 
-    // First, we need to scan and ensure that the Configurator's requested BSS is
-    // within range for us / exists.
-
-    // TODO: scan!
-
-
-    // TODO: associate with bSTA config information from Configurator!
-
-    ec_status_code_t connection_status = DPP_STATUS_CONFIG_REJECTED;
-
-
-    // Only necessary if Configurator includes "sendConnStatus" in Configuration response.
-    bool needs_connection_status = (send_connection_status_attr.has_value());
+    // First, we immediately send a Connection Status Result frame to Configurator per
+    // EasyConnect 6.4.4 after parsing/ensuring the Configuration Response is valid
 
     // Mandatory Configuration Result frame indicating configuration status.
+    // The spec is a bit goofy here and says that DPP Status should indicate
+    // the success or failure of _applying the configuration_
+    // It does not mention that this status indicates real association success/failure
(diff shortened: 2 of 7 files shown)


src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -1297,5 +1278,61 @@ bool ec_enrollee_t::add_presence_announcement_freq(unsigned int freq)
...
+        case wifi_connection_status_disabled:
+            dpp_status = DPP_STATUS_CONFIGURATION_FAILURE;
+            break;
+        case wifi_connection_status_disconnected:
+            dpp_status = DPP_STATUS_CONFIG_REJECTED;
+            break;
+        default:
+            em_printfout("Unhandled Wi-Fi connection status enum %d", sta_data.stats.connect_status);
+            return false;
+    }
```
bcarlson-dev, line 1311: So unfortunately based on the granularity of data we get from OneWifi it seems like we might not be able to ever return `STATUS_AUTH_FAILURE` (as you correctly do not have) however seems like the cases for `DPP_STATUS_CONFIGURATION_FAILURE` and `DPP_STATUS_CONFIG_REJECTED` are not correct here.

It seems like there is more than meets the eye in the phrase **Immediately after reaching any status indicated in Table 22** which could mean that the `DPP_STATUS_CONFIGURATION_FAILURE` and `DPP_STATUS_CONFIG_REJECTED` are sent sooner than after an association attempt? The spec only shows positive cases so it's hard to nail down. Either way I think we can get away with sending that frame here but we have to do the check against the connectors _somehow_.

image

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -1297,5 +1278,61 @@ bool ec_enrollee_t::add_presence_announcement_freq(unsigned int freq)
...
+
+    // Did the Configurator request a Connection Status Result Frame?
+    if (!m_configurator_requested_conn_status_frame) {
+        return true;
+    }
```
bcarlson-dev, line 1316: Is this variable really needed? Can't we just add the BSSID to `m_awaiting_assoc_status` **iff** the `sendConnStatus` attrib is present, making the original check at the top all that is needed?

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -1297,5 +1278,61 @@ bool ec_enrollee_t::add_presence_announcement_freq(unsigned int freq)
...
+
+    auto [frame, frame_len] = create_connection_status_result(dpp_status, std::string(sta_data.bss_info.ssid));
+    if (frame == nullptr || frame_len == 0) {
+        em_printfout("Failed to create Configuration Connection Status Result frame!");
+        return false;
+    }
+
+    bool sent = m_send_action_frame(const_cast<uint8_t*>(m_configurator_mac.data()), frame, frame_len, m_selected_freq, 0);
```
bcarlson-dev, line 1324: I don't love the `m_configurator_mac` variable since it kind of introduces this singular nature, especially when you're accounting for possibly multiple awaiting responses.

Could we make `m_awaiting_assoc_status` a `BSSID -> PA/Sender MAC` map so we could get this MAC from the map when we index it?

bcarlson-dev (changes requested):

src/em/prov/easyconnect/ec_enrollee.cpp
tuckerpo-cl: Yeah I think `DPP_STATUS_CONFIGURATION_FAILURE` and `DPP_STATUS_CONFIG_REJECTED` are meant to be the status for the Configuration Result frame (not this Configuration Connection Status Result frame). Also, for `STATUS_INVALID_CONNECTOR` and `STATUS_NO_MATCH`, those assume you're mid-DPP Peer Exchange/network introduction, which we're not currently doing.

Regarding translating the OneWifi `wifi_connection_status_t` to DPP status codes, it's clear that `wifi_connection_status_connected` maps to `DPP_STATUS_OK`, `wifi_connection_status_ap_not_found` maps to `DPP_STATUS_NO_AP_DISCOVERED`, but that leaves the remaining two states (`wifi_connection_status_disconnected` and `wifi_connection_status_disabled`) sort of in limbo since there's no clean mapping according to table 22 🤔

bcarlson-dev: I read in the `rdk-wifi-hal` and with regards to association it only sets `wifi_connection_status_connected` and if `!= OK` then `wifi_connection_status_ap_not_found`. All those other statuses are set elsewhere so for our purposes there really isn't any match. We can lump them all in `NO_AP` if you want just for completeness but I don't expect those to actually be set as the first call on that callback for that BSSID.

tuckerpo-cl: Yeah I agree that `NO_AP` is probably the most "generic" status that we could map those to, I'll do that

inc/ec_enrollee.h
```diff
@@ -12,6 +12,7 @@
 #include <unordered_set>
 #include <atomic>
 #include <thread>
+#include <array>
```
tuckerpo-cl, line 15: Whoops, no longer needed

bcarlson-dev (approved):

PR #331, opened 2025-04-24
Connect bSTA to BSS and onboard onto EasyMesh network when DPP Configuration Response is recieved

bcarlson-dev:
Depends on the merging of:
- `OneWifi` [PR #284](https://github.com/rdkcentral/OneWifi/pull/284) ✅ 
- `OneWifi` [PR #290](https://github.com/rdkcentral/OneWifi/pull/290) ✅ 
-  `rdk-wifi-hal` [PR #160](https://github.com/rdkcentral/rdk-wifi-hal/pull/160) ✅

### DIFF
--- a/inc/ec_configurator.h
+++ b/inc/ec_configurator.h
@@ -69,7 +69,7 @@ using start_stop_clist_build_func = std::function<bool(bool)>;
  * @param bssid The BSSID of the BSS to connect to
  * @return bool true if the attempt was performed successfully, false otherwise
  */
-using bsta_connect_func = std::function<bool(const std::string&, const std::string, bssid_t)>;
+using bsta_connect_func = std::function<bool(const std::string&, const std::string&, bssid_t)>;
 
 /**
  * @brief Creates a DPP Configuration Response object for the backhaul STA interface.
--- a/inc/ec_configurator.h
+++ b/inc/ec_configurator.h
@@ -51,6 +51,27 @@ using send_act_frame_func = std::function<bool(uint8_t*, uint8_t *, size_t, unsi
 using toggle_cce_func = std::function<bool(bool)>;
 
 /**
+ * @brief Initial setup for starting or stopping the building of the channel list
+ * 
+ * @param do_start true to start the process, false to stop it
+ * @return bool true if the action was successful, false otherwise
+ * 
+ * @note This does not build the channel list but just sets up the environment for the process
+ */
+using start_stop_clist_build_func = std::function<bool(bool)>;
+
+
+/**
+ * @brief Attempts a connection between the backhaul STA to the specified BSS.
+ * 
+ * @param ssid The SSID of the BSS to connect to
+ * @param passphrase The passphrase for the BSS
+ * @param bssid The BSSID of the BSS to connect to
+ * @return bool true if the attempt was performed successfully, false otherwise
+ */
+using bsta_connect_func = std::function<bool(const std::string&, const std::string, bssid_t)>;
+
+/**
  * @brief Creates a DPP Configuration Response object for the backhaul STA interface.
  * @param conn_ctx Optional connection context (not needed for Enrollee, needed for Configurator) -- pass nullptr if not needed.
  * @return cJSON * on success, nullptr otherwise
--- a/src/agent/em_agent.cpp
+++ b/src/agent/em_agent.cpp
@@ -858,10 +858,10 @@ bool em_agent_t::set_disconnected_steady_state()
     wifi_bus_desc_t *desc = get_bus_descriptor();
     ASSERT_NOT_NULL(desc, false, "%s:%d descriptor is null\n", __func__, __LINE__);
 
-    raw_data_t raw_act_frame;
-    memset(&raw_act_frame, 0, sizeof(raw_data_t));
-    raw_act_frame.data_type = bus_data_type_none;
-    if (desc->bus_set_fn(&m_bus_hdl, WIFI_SET_DISCONN_STEADY_STATE, &raw_act_frame)== 0) {
+    raw_data_t empty_arg;
+    memset(&empty_arg, 0, sizeof(raw_data_t));
+    empty_arg.data_type = bus_data_type_none;
+    if (desc->bus_set_fn(&m_bus_hdl, WIFI_SET_DISCONN_STEADY_STATE, &empty_arg)== 0) {
         em_printfout("Set Disconnected Steady State succeeded");
         return true;
     }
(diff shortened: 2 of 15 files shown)


inc/ec_configurator.h
```diff
@@ -50,6 +50,27 @@ using send_act_frame_func = std::function<bool(uint8_t*, uint8_t *, size_t, unsi
...
+
+/**
+ * @brief Attempts a connection between the backhaul STA to the specified BSS.
+ * 
+ * @param ssid The SSID of the BSS to connect to
+ * @param passphrase The passphrase for the BSS
+ * @param bssid The BSSID of the BSS to connect to
+ * @return bool true if the attempt was performed successfully, false otherwise
+ */
+using bsta_connect_func = std::function<bool(const std::string&, const std::string, bssid_t)>;
```
tuckerpo-cl, line 72: I think `passphrase` can be passed by const-ref here (arg 2)

inc/em_base.h
```diff
@@ -86,7 +86,7 @@ extern "C"
 
 #define   EM_MAX_EVENT_DATA_LEN   4096*100
 #define EM_MAX_CHANNELS_IN_LIST  64
-#define EM_MAX_CMD_GEN_TTL  10
+#define EM_MAX_CMD_GEN_TTL  15
```
tuckerpo-cl, line 89: Just for my own knowledge is this like 15 retries? 15 second timeout?

src/agent/em_agent.cpp
```diff
@@ -852,6 +852,40 @@ bool em_agent_t::send_action_frame(uint8_t dest_mac[ETH_ALEN], uint8_t *action_f
...
 
+bool em_agent_t::set_disconnected_steady_state()
+{
+    
+    wifi_bus_desc_t *desc = get_bus_descriptor();
+    ASSERT_NOT_NULL(desc, false, "%s:%d descriptor is null\n", __func__, __LINE__);
+
+    raw_data_t raw_act_frame;
+    memset(&raw_act_frame, 0, sizeof(raw_data_t));
+    raw_act_frame.data_type = bus_data_type_none;
```
tuckerpo-cl, line 863: This is probably just a copy+paste relic but could we call `raw_data_t` here something like `raw_data = {0};`? The current naming suggests it's crafting/sending an action frame.

src/agent/em_agent.cpp
```diff
@@ -852,6 +852,40 @@ bool em_agent_t::send_action_frame(uint8_t dest_mac[ETH_ALEN], uint8_t *action_f
...
+
+bool em_agent_t::set_disconnected_scan_none_state()
+{
+    
+    wifi_bus_desc_t *desc = get_bus_descriptor();
+    ASSERT_NOT_NULL(desc, false, "%s:%d descriptor is null\n", __func__, __LINE__);
+
+    raw_data_t raw_act_frame;
+    memset(&raw_act_frame, 0, sizeof(raw_data_t));
+    raw_act_frame.data_type = bus_data_type_none;
```
tuckerpo-cl, line 880: Same `raw_act_frame` naming comment as above

src/em/em.cpp
```diff
@@ -645,30 +645,32 @@ int em_t::send_frame(unsigned char *buff, unsigned int len, bool multicast)
...
-    bool is_colocated = (memcmp(ctrl_al, agent_al, ETH_ALEN) == 0);
+    auto ctrl_em_al = m_data_model->get_controller_interface_mac();
+    auto agent_em_al = m_data_model->get_agent_al_interface_mac();
+    bool is_colocated = (memcmp(ctrl_em_al, agent_em_al, ETH_ALEN) == 0);
 
     AlServiceDataUnit sdu;
     sdu.setSourceAlMacAddress(g_al_mac_sap);
     if (m_service_type == em_service_type_ctrl) {
-        if (is_colocated) {
+        if (is_colocated && is_loopback_node) {
```
tuckerpo-cl, line 666: Should this be `||`?

src/em/prov/em_provisioning.cpp
```diff
@@ -605,9 +605,13 @@ cJSON *em_provisioning_t::create_enrollee_bsta_list(ec_connection_context_t *con
...
-    for (unsigned int i = 0; i < dm->get_num_radios(); i++) {
-        dm_radio_t *radio = dm->get_radio(i);
-        if (!radio) {
+    for (unsigned int i = 0; i < dm->get_num_bss(); i++) {
+        em_bss_info_t *bss_info = dm->get_bss_info(i);
+        if (!bss_info) {
+            continue;
+        }
+        // Skip if not backhaul
+        if (bss_info->id.haul_type != em_haul_type_backhaul) {
```
tuckerpo-cl, line 614: I'm fine with this change although the EasyMesh spec (if I recall correctly) calls for _all_ radios on the Enrollee to be reported in the Configuration Request object. Again, I'm fine with just backhaul in this impl, but might not be spec compliant (or I could just be wrong!).

src/orch/em_orch_agent.cpp
```diff
@@ -49,6 +50,19 @@ void em_orch_agent_t::orch_transient(em_cmd_t *pcmd, em_t *em)
...
+        auto agent_int = pcmd->get_agent_al_interface();
+        std::string al_mac_key = util::mac_to_string(agent_int->mac) + "_al";
+        em_t* al_node = (em_t*)hash_map_get(m_mgr->m_em_map, al_mac_key.c_str());
+        if (al_node != NULL && al_node->m_ec_manager && al_node->m_ec_manager->is_enrollee_onboarding()) {
+            // If the enrollee is still onboarding, we need to wait for it to finish before timing out
+            // Lets reset the timeout
+            gettimeofday(&pcmd->m_start_time, NULL);
+            stats->time = 0;
+        }
+    }
```
tuckerpo-cl, line 64: It would be nice if this could somehow be event driven, i.e. on Enrollee association success, restart `em_cmd_type_dev_init` as opposed to "polling" -- but totally fine with this if it works

tuckerpo-cl (changes requested):
Mostly nit-picks but overall LGTM

inc/ec_configurator.h
bcarlson-dev: Good catch

inc/em_base.h
bcarlson-dev: This may no longer be needed since I changed by approach but it was adjusting the timeout

src/agent/em_agent.cpp
bcarlson-dev: Totally fair

src/orch/em_orch_agent.cpp
bcarlson-dev: While I would prefer that as well, I couldn't really see a way for that to happen without really mucking up the code. The current code is supposed to poll if `time_running < timeout` so cleanest way was to just reset `time_running`. I did spend considerable time looking for an event driven alternative but I really couldn't find a viable option

src/em/em.cpp
bcarlson-dev: So, your comment got me thinking and while the OR wouldn't necessarily matter I think the whole `is_colocated` can just be removed because if it's co-located then it's a loopback node and if it's not colocated but sending to itself (don't know when this would be) then it's a loop back node.

src/em/prov/em_provisioning.cpp
bcarlson-dev: So I thought so as well but then I looked at R5 again and thought that maybe it's just sending the **one** it will be using for the backhaul STA and for R6 it's the MLD radios which is why we were confused. I could also be mixed up again for that

tuckerpo-cl (approved):

PR #333, opened 2025-04-24
Collecting assoc sta stats as part of Ap metrics response per "Metrics reporting policy"!

Nikita-Hakai:
Collecting assoc sta stats as part of Ap metrics response!

Test procedure: Trigger sta link metrics and traffic stats policy using CLI while connecting/disconnecting clients to the AP. If connected and policy is set, stats collection can be seen updated over Client connections menu.

### DIFF
--- a/inc/em_base.h
+++ b/inc/em_base.h
@@ -1952,6 +1952,8 @@ typedef enum {
     em_cmd_type_get_mld_config,
     em_cmd_type_mld_reconfig,
     em_cmd_type_beacon_report,
+    em_cmd_type_ap_metrics_report,
+
     em_cmd_type_max,
 } em_cmd_type_t;
 
@@ -2597,6 +2599,7 @@ typedef enum {
     em_bus_event_type_get_sta_client_type,
     em_bus_event_type_cce_ie,
     em_bus_event_type_assoc_status,
+    em_bus_event_type_ap_metrics_report,
 
     em_bus_event_type_max
 } em_bus_event_type_t;
@@ -2825,6 +2828,13 @@ typedef struct {
     em_disassoc_params_t	params[MAX_STA_TO_DISASSOC];
 } em_cmd_disassoc_params_t;
 
+typedef struct {
+    mac_address_t   ruid;
+    bool sta_link_metrics_include;
+    bool sta_traffic_stats_include;
+    bool wifi6_status_report_include;
+} __attribute__((__packed__)) em_cmd_ap_metrics_rprt_params_t;
+
 typedef enum {
     em_network_node_data_type_invalid,
     em_network_node_data_type_false,
@@ -2864,6 +2874,7 @@ typedef struct {
         em_cmd_btm_report_params_t  btm_report_params;
         em_cmd_disassoc_params_t	disassoc_params;
 		em_cmd_scan_params_t	scan_params;
+        em_cmd_ap_metrics_rprt_params_t ap_metrics_params;
     } u;
 	em_network_node_t *net_node;
 } em_cmd_params_t;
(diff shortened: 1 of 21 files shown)


src/ctrl/em_ctrl.cpp
```diff
@@ -643,9 +643,10 @@ em_t *em_ctrl_t::find_em_for_msg_type(unsigned char *data, unsigned int len, em_
             }
             break;
 
-        case em_msg_type_topo_notif:
+        /* case em_msg_type_topo_notif:
```
amarnathhullur, line 646: This comment would comment out the whole code till 684. Is that the intent for this code?

inc/em_base.h
```diff
@@ -914,9 +914,10 @@ typedef struct {
 
 typedef struct {
     mac_address_t sta_mac;
+    bssid_t bssid;
     em_string_t sta_client_type;
-    unsigned char num_bssids;
-    em_assoc_vendor_link_metrics_t assoc_vendor_link_metrics[0];
+    //unsigned char num_bssids;
+    //em_assoc_vendor_link_metrics_t assoc_vendor_link_metrics[0];
```
amarnathhullur, line 920: Why is this commented? Is this not needed?

src/agent/dm_easy_mesh_agent.cpp
```diff
@@ -738,6 +736,73 @@ int dm_easy_mesh_agent_t::analyze_beacon_report(em_bus_event_t *evt, em_cmd_t *p
...
+
+    if (strstr((const char *)evt->u.raw_buff, "Associated Wi-Fi 6 STA Status Report") != NULL) {
+        //printf("Associated Wi-Fi 6 STA Status Report found in JSON.\n");
+        evt_param->u.ap_metrics_params.wifi6_status_report_include = true;
+    } else {
+        //printf("Associated Wi-Fi 6 STA Status Report not found in JSON.\n");
+        evt_param->u.ap_metrics_params.wifi6_status_report_include = false;
+    }
+
+    pcmd[num] = new em_cmd_ap_metrics_report_t(evt->params, *(this));
```
amarnathhullur, line 794: How will this object get freed?

src/agent/dm_easy_mesh_agent.cpp
```diff
@@ -738,6 +736,73 @@ int dm_easy_mesh_agent_t::analyze_beacon_report(em_bus_event_t *evt, em_cmd_t *p
...
+    tmp = pcmd[num];
+    num++;
+
+    while ((pcmd[num] = tmp->clone_for_next()) != NULL) {
```
amarnathhullur, line 798: What is this line addressing?

src/ctrl/em_ctrl.cpp
Nikita-Hakai: This was a temporary fix for multiple bssid entries issue, because of which DM selection was wrong at times. Reverting the change as it is now fixed by Aswini.

inc/em_base.h
Nikita-Hakai: This was custom tlv created earlier to collect stats which is not used now, need to confirm with Soumya after demo if this can be removed.

src/agent/dm_easy_mesh_agent.cpp
Nikita-Hakai: When the command is destroyed as part of succesful command orchestration, if not if would get desrtoyed after a timeout occurs.

src/agent/dm_easy_mesh_agent.cpp
Nikita-Hakai: It would clone the command in case we created more than one. In this case, it is only one command.

amarnathhullur (approved):

PR #342, opened 2025-04-29
Fixed crash issue with controller

prashant-singh1:
Reason for change: 1) Multiple client reports are causing the same STA info to be repeatedly updated. This is followed by a repeated DB update, after which the entry is removed from the Assoc map again repeatedly, leading to the crash.
Test Procedure: With this changes we observed controller is running without crash.
Risks: Medium
Priority: P1


amarnathhullur (approved):
Looks good to me

amarnathhullur (approved):

PR #345, opened 2025-04-30
Separate data socket and control socket for al_sap lib

mzalfres-mobica:
al__sap library change to implement two separate sockets, for data and control channel.

### DIFF
--- a/inc/al_service_access_point.h
+++ b/inc/al_service_access_point.h
@@ -27,7 +27,7 @@ public:
 	 *
 	 * @note Ensure the socket path is valid and accessible.
 	 */
-	AlServiceAccessPoint(const std::string& socketPath);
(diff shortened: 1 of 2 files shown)


mzalfres-mobica:
> The two socket system is not implemented by the C++ tunnel and is only used by the Rust 1905 daemon at the moment. Given that the demos are less than a week away and the Rust 1905 daemon is not going to be used (to our knowledge), I don't think this change should be merged until after the demos.

Completely agree. Please let us know when we can continue. So far we'll work on our forked repos.

amarnathhullur:
Looks fine to me. What is the reason to have two sockets control and data? Who will be using control socket and who will be using data sockets?

torrentius (approved):
It is ok!!

PR #356, opened 2025-05-10
Implements DPP Reconfiguration Announcement logic + channel list building

tuckerpo-cl:
Depends on https://github.com/rdkcentral/OneWifi/pull/324 for dynamic channel list

Still TODO is determining when to begin the Reconfiguration flow (and by proxy when to actually start the Reconfiguration Announcement thread included in this PR)

### DIFF
--- a/inc/ec_enrollee.h
+++ b/inc/ec_enrollee.h
@@ -130,6 +130,16 @@ public:
 	 */
 	bool handle_gas_initial_response(ec_gas_initial_response_frame_t *frame, size_t len, uint8_t src_mac[ETH_ALEN]);
 
+	/**
+	 * @brief Handle a Reconfiguration Authentication Request frame
+	 * 
+	 * @param frame The Reconfiguration Authentication Reuqest frame
+	 * @param len The length of the frame
+	 * @param src_mac Where the frame came from (Configurator)
+	 * @return true on success, otherwise false
+	 */
+	bool handle_recfg_auth_request(ec_frame_t *frame, size_t len, uint8_t src_mac[ETH_ALEN]);
+
     
 	/**!
 	 * @brief Tears down the existing connection by freeing associated resources.
@@ -195,6 +205,21 @@ public:
 	 */
 	bool handle_assoc_status(const rdk_sta_data_t &sta_data);
 
+	/**
+	 * @brief Handle a BSS info event
+	 * 
+	 * This callback is triggered by OneWifi scanning
+	 * 
+	 * If this BSS info event contains the SSID that we were instructed to 
+	 * connect to (in the Configuration Response from the Configurator), we add
+	 * this BSS info's heard frequency to our Reconfiguration Announcement 
+	 * broadcast frequency list
+	 * 
+	 * @param bss_info The BSS info containing SSID and frequency
+	 * @return true on success, otherwise false
+	 */
+	bool handle_bss_info_event(const wifi_bss_info_t &bss_info);
+
 private:
 
     
@@ -208,6 +233,14 @@ private:
 	 * @note See: EasyConnect 6.2 DPP Presence Announcement
 	 */
 	void send_presence_announcement_frames();
+
+	/**
+	 * @brief Sends DPP Reconfiguration Announcement frames until a DPP Reconfiguration Authentication Request frame is received
+	 * 
+	 * 
+	 * @note See: EasyConnect 6.5.2 DPP Reconfiguration Announcement
+	 */
+	void send_reconfiguration_announcement_frames();
 
     std::string m_mac_addr;
 
@@ -417,6 +450,19 @@ private:
      */
     std::unordered_set<uint32_t> m_pres_announcement_freqs = {2437, 5220};
 
+	/**
+	 * @brief Set of frequencies to broadcast Reconfiguration Announcement frames on
+	 * 
+	 * This list should be extended as follows (EasyConnect 6.5.2):
+	 * 
+	 * 1. If a beacon/probe response is heard advertising the SSID we were given in the Configuration response object,
+	 * add the channel where this frame was heard
+	 * 
+	 * 2. Channels where a CCE IE was heard
+	 * 
+	 */
+	std::unordered_set<uint32_t> m_recnf_announcement_freqs = {2437, 5220};
+
     /**
      * @brief The frequency to send action frames on
      * 
@@ -432,6 +478,13 @@ private:
      */
     std::atomic<bool> m_received_auth_frame{false};
 
+	/**
+	 * @brief True if we've received DPP Reconfiguration Authentication Request frame
+	 * 
+	 * Signals that this Enrollee should stop sending Reconfiguration Announcement frames
+	 */
+	std::atomic<bool> m_received_recfg_auth_frame{false};
+
     /**
      * @brief Thread for sending DPP Presence Announcement frames upon onboarding start
      * 
@@ -443,6 +496,17 @@ private:
      * our DPP Authentication Frame, and keep Presence Announcing forever
      */
     std::thread m_send_pres_announcement_thread;
+
+	/**
+	 * @brief Thread for sending Reconfiguration Announcement frames
+	 * 
+	 * Sends Reconfig Announcement frames until a Reconfiguration Authentication Request frame is heard
+	 * 
+	 * Must be on a seperate thread so as not to block the receiving
+	 * of the Reconfiguration Authentication Request frame
+	 * 
+	 */
+	std::thread m_send_recfg_announcement_thread;
 
 	/**
 	 * @brief Metadata about a fragmented GAS response frame
@@ -468,6 +532,17 @@ private:
 	std::unordered_map<std::string, std::vector<uint8_t>> m_awaiting_assoc_status = {};
 
 	bool m_is_onboarding = false;
+
+	/**
+	 * @brief The SSID contained in the Configuration Response that we're meant to connect to
+	 * 
+	 * This is used in part to rebuild a channel list for DPP Reconfiguration
+	 * 
+	 * EasyConnect 6.5.2:
+	 * 	For each channel on which the Enrollee detects the SSID for which it is currently configured, add to the channel list
+	 * 
+	 */
+	std::string m_configured_ssid = {};
 };
 
 #endif // EC_ENROLLEE_H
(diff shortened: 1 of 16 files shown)


src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -566,6 +612,8 @@ bool ec_enrollee_t::handle_config_response(uint8_t *buff, size_t len, uint8_t sa
         return false;
     }
 
+    m_configured_ssid = std::string(bsta_ssid->valuestring);
```
bcarlson-dev, line 615: Possibly set this on successful association instead of just a configuration response

inc/ec_enrollee.h
```diff
@@ -468,6 +532,17 @@ class ec_enrollee_t {
 	std::unordered_map<std::string, std::vector<uint8_t>> m_awaiting_assoc_status = {};
 
 	bool m_is_onboarding = false;
+
+	/**
+	 * @brief The SSID contianed in the Configuration Response that we're meant to connect to
```
bcarlson-dev, line 537:
```suggestion
	 * @brief The SSID contained in the Configuration Response that we're meant to connect to
```
Typo

bcarlson-dev (changes requested):
Some small nitpicks although I want to throughly check stuff out when I get home

src/em/prov/easyconnect/ec_enrollee.cpp
tuckerpo-cl: Yeah from a business logic POV that makes sense, but the spec(s) call it out as whatever SSID was parsed from the most recently received Configuration Response :shrug:

tuckerpo-cl: This is not explicitly said but sort of implied in EC 6.5.2 ss 2 `For each channel on which the Enrollee detects the SSID for which it is currently configured, add to the channel list;`. I guess the interpretattion of "currently configured" is open to the implementation.

It could either be:

1. The most recently received Configuration Response object's SSID, which is how my code interprets it, similar to how we immediately send a Connection Status frame **upon receiving the Configuration Response** instead of waiting for association to occur
2. Use the SSID the Enrollee was most recently successfully able to associate to

🤷

tuckerpo-cl:
@bcarlson-dev Another interesting piece is both EC and EM specs say that it's the Enrollee's burden to know when to begin Reconfiguration, which is a bit goofy IMO.

If the Enrollee bSTA is disassociated or disconnected, we (obviously) won't receive a Topology Notification from the Controller telling us that the Enrollee has left, which IMO would be the easiest way (in an EasyMesh implementation) to kick things off, but of course that's not possible if the bSTA link is down.

You could also have the Controller/Configurator notify the Enrollee that it has disconnected via a VS field in a GAS frame, but again, both specs say that the **Enrollee** must kick off re-conf.

From EC 2.6.2:
```
2.6.2 Triggers for Reconfiguration
There can be multiple reasons for an Enrollee to reinitiate a connection with its Configurator, for example, the Enrollee's
Connector may have expired and needs renewal. Alternately, a PSK or SAE password may have expired or changed.
Depending on the credential being used, an enrollee may determine the need for reconfiguration at different stages in its
network access lifecycle.
• WPA2 PSK/Passphrase - A STA using a WPA2 PSK or Passphrase follows the standard IEEE 802.11
procedures and an unsuccessful 4-way handshake using the provided PSK or Passphrase may trigger
reconfiguration
• WPA3 Password - A STA using a WPA3 Personal password follows the SAE authentication as per the IEEE
802.11 procedures and an unsuccessful SAE authentication may trigger reconfiguration
• DPP Connector - A STA using a DPP Connector attempts the Network Introduction Protocol as specified in
Section 6.6 and a failure with error STATUS_INVALID_CONNECTOR may trigger reconfiguration
```

And 5.3.10.2 from EM:

```
As part of a backhaul BSS reconfiguration, backhaul STA connectivity of some of the Multi-AP Agents in the Multi-AP
network may be lost. If a backhaul STA of a Multi-AP Agent loses connectivity to its backhaul BSS or receives a protected
Disassociation frame, it shall try to re-associate its backhaul STA using the backhaul BSS credentials from the last
onboarding procedure. If the re-association fails and the Multi-AP Agent possesses a Privacy-protection-key from the
Multi-AP Controller, then the Multi-AP Agent shall invoke the reconfiguration algorithm in section 6.5.2 of [18] by starting
the Reconfiguration Announcement procedures
```

So, fairly poorly defined in both specs. We'll need to write some sort of all encompassing "is my bSTA link still OK?" daemon thread on the Enrollee I imagine.

bcarlson-dev:
> @bcarlson-dev Another interesting piece is both EC and EM specs say that it's the Enrollee's burden to know when to begin Reconfiguration, which is a bit goofy IMO.
> 
> If the Enrollee bSTA is disassociated or disconnected, we (obviously) won't receive a Topology Notification from the Controller telling us that the Enrollee has left, which IMO would be the easiest way (in an EasyMesh implementation) to kick things off, but of course that's not possible if the bSTA link is down.
> 
> You could also have the Controller/Configurator notify the Enrollee that it has disconnected via a VS field in a GAS frame, but again, both specs say that the **Enrollee** must kick off re-conf.
> 
> From EC 2.6.2:
> 
> ```
> 2.6.2 Triggers for Reconfiguration
> There can be multiple reasons for an Enrollee to reinitiate a connection with its Configurator, for example, the Enrollee's
> Connector may have expired and needs renewal. Alternately, a PSK or SAE password may have expired or changed.
> Depending on the credential being used, an enrollee may determine the need for reconfiguration at different stages in its
> network access lifecycle.
> • WPA2 PSK/Passphrase - A STA using a WPA2 PSK or Passphrase follows the standard IEEE 802.11
> procedures and an unsuccessful 4-way handshake using the provided PSK or Passphrase may trigger
> reconfiguration
> • WPA3 Password - A STA using a WPA3 Personal password follows the SAE authentication as per the IEEE
> 802.11 procedures and an unsuccessful SAE authentication may trigger reconfiguration
> • DPP Connector - A STA using a DPP Connector attempts the Network Introduction Protocol as specified in
> Section 6.6 and a failure with error STATUS_INVALID_CONNECTOR may trigger reconfiguration
> ```
> 
> And 5.3.10.2 from EM:
> 
> ```
> As part of a backhaul BSS reconfiguration, backhaul STA connectivity of some of the Multi-AP Agents in the Multi-AP
> network may be lost. If a backhaul STA of a Multi-AP Agent loses connectivity to its backhaul BSS or receives a protected
> Disassociation frame, it shall try to re-associate its backhaul STA using the backhaul BSS credentials from the last
> onboarding procedure. If the re-association fails and the Multi-AP Agent possesses a Privacy-protection-key from the
> Multi-AP Controller, then the Multi-AP Agent shall invoke the reconfiguration algorithm in section 6.5.2 of [18] by starting
> the Reconfiguration Announcement procedures
> ```
> 
> So, fairly poorly defined in both specs. We'll need to write some sort of all encompassing "is my bSTA link still OK / have my credentials expired / did association fail / do I still have connectivity?" daemon thread on the Enrollee I imagine.

My primary thought, at least from a "I plug in an already onboarded extender and it can't onboard anymore" is that once that onboarding fails it would just trigger reauth/reconfig instead of auth/config. As I'm writing this, it's at the lowest level "can my bSTA associate to the BSS I've been configured to join". If this fails (and we were previously authenticated) then we should just start reconfig. 

With regards to where to place the code to kick off the thread. In `ec_enrollee::start_onboarding` there is a `do_reconfig` parameter (currently only false) which is currently unused beyond some key-specific things. That method is also where you do the `m_send_pres_announcement_thread` instantiation so I think an if-else can just be placed there. 

There will need to be a little bit of restructuring of ideas on when to run `start_onboarding` and maybe adding some logic in `ec_manager_t::handle_assoc_status` to determine when to do the reconfiguration (then just delete the enrollee and re-instantiate it, starting it with `do_reconfig=true`) but these are more through discussions that will need to be had outside of this PR.

src/em/prov/easyconnect/ec_pa_configurator.cpp
```diff
@@ -266,11 +266,7 @@ bool ec_pa_configurator_t::process_proxy_encap_dpp_msg(em_encap_dpp_t *encap_tlv
             break;
         }
         case ec_frame_type_recfg_auth_req: {
-            em_printfout("Encap DPP frame type (%d) not handled", ec_frame_type);
-            std::vector<uint8_t> encap_frame_vec(encap_frame, encap_frame + encap_frame_len);
-            // Will be compared against incoming presence announcement hash and mac-addr
-            m_stored_recfg_auth_frames.push_back(encap_frame_vec); 
-            did_finish = true;
+            did_finish = m_send_action_frame(dest_mac, encap_frame, encap_frame_len, 0, 0);
```
bcarlson-dev, line 269: I'm a bit confused by this removal. Based on my reading of the EasyMesh spec, reconfig does the same thing as auth where:

1. PA Recieves Presence 
2. Send Hash to Controller
3.  Controller Creates (Reconfig) Auth Request 
4. Controller encaps and sends (Reconfig) Auth Request to PA 
5. PA caches (Reconfig) auth request mapped to hash and waits

With a later on:

1.  PA Recieves Presence
2. PA reads cached (Reconfig) Auth Request
3. PA sends frame


EasyMesh 5.3.10.2

> If a Proxy Agent receives from the Multi-AP Controller a Proxied Encap DPP message that includes an encapsulated DPP Reconfiguration Authentication Request frame destined for a Multi-AP Agent seeking reconfiguration, the Proxy Agent shall listen for a DPP Reconfiguration Announcement frame. If a Proxy Agent receives a DPP Reconfiguration Announcement frame, it shall check that the hash in the frame matches the hash of the Controller's C-sign-key. If the hash matches and the DPP Reconfiguration Authentication Request frame is for the Multi-AP Agent seeking reconfiguration based on the matching of the source MAC address of the Reconfig Announcement frame with the Destination STA MAC Address field from the 1905 Encap DPP TLV, the Proxy Agent shall send the DPP Reconfiguration Authentication Request frame to the Multi-AP Agent seeking reconfiguration. If the hash does not match, the Proxy Agent shall forward the Reconfiguration Announcement frame to the Multi-AP Controller as specified above.

tuckerpo-cl: Ah, you're right. I'll fix that

bcarlson-dev (approved):

PR #358, opened 2025-05-13
RDKBACCL-578: EasyMesh Agent integration in BPI Extender device

ap934:
Reason for change: Added EM_EXTENDER flag for bpi extender builds
Test procedure: Build should go through, and only agent should get compiled in extender
Risks: None


manigandanj:
looks like configure.ac change is missing.

PR #365, opened 2025-05-27
Issue #347 Push L1 unit tests for 7headers

rtaruni:

PR #370, opened 2025-06-02
Change `toggle_cce` to only add/remove to backhaul BSSs and to refresh the correct bands

bcarlson-dev:
* Corrected a logic error in `pa_cfg_toggle_cce` within `ec_manager_t` by removing an incorrect negation in the condition that checks if the instance is a controller or if the configurator is null.
* Added checks in `toggle_cce` to verify that BSSs exist, are enabled, and are backhaul BSSs before attempting to add or remove DPP IEs. 
* Introduced logic to track frequency bands of BSSs when toggling CCEs allowing OneWifi to refresh the correct BSSs.

Depends on [OneWifi #365](https://github.com/rdkcentral/OneWifi/pull/365) to work

### DIFF
--- a/src/em/em.cpp
+++ b/src/em/em.cpp
@@ -723,48 +723,88 @@ bool em_t::toggle_cce(bool enable)
 {
     const unsigned int num_bss = m_data_model->get_num_bss();
 
-    bool success = false;
-    if (enable) {
-        printf("Adding DPP IE to %d BSSs\n", num_bss);
-        for (unsigned int i = 0; i < num_bss; i++) {
-            dm_bss_t* bss = m_data_model->get_bss(i);
-            em_bss_info_t* bss_info = bss->get_bss_info();
-            em_interface_t* bssid = &bss_info->bssid;
-
-            success = bss->add_vendor_ie(&ec_manager_t::CCE_IE);
-            if (!success) {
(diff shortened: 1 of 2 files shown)


src/em/em.cpp
```diff
@@ -723,48 +723,88 @@ bool em_t::toggle_cce(bool enable)
...
+    // Refresh OneWifi for each band
+    for (const auto& band : bands) {
+        webconfig_subdoc_type_t vap_type = dm_easy_mesh_t::get_subdoc_vap_type_for_freq(band);
+        printf("Refreshing OneWifi for band %d with subdoc type %d\n", band, vap_type);
+        int refresh_outcome = m_mgr->refresh_onewifi_subdoc("'Vendor IE Refresh'", vap_type);
+        if (refresh_outcome != 1) {
+            printf("Error occurred on Vendor IE Refresh: return value %d\n", refresh_outcome);
+            return false;
+        }
+    }
```
tuckerpo-cl, line 806: Should this be inside an `if (success)` scope? It seems if success is false, this is unnecessary work

tuckerpo-cl (approved):
LGTM mod 1 nit

tuckerpo-cl (approved):

PR #377, opened 2025-06-05
Recfg Auth Response / Recfg Auth Confirm

tuckerpo-cl:
1. Configurator implementation of `handle_recfg_auth_response`
2. Configurator creation and sending of Reconfiguration Authentication Confirm frames
3. Enrollee handling of Reconfiguration Authentication Confirm frames in `handle_recfg_auth_confirm`
4. Enrollee creation and sending of Configuration Request frame once Reconfiguration Authentication Confirm frame is handled
5. Fixes bug related to storing / parsing of Connectors as attributes: both Configurator and Enrollee store their connectors in their x.y.z base64 encoded signed form, and we were parsing them as if they were JSON.

The spec poorly defines what should happen on the Enrollee if any step of handling the Reconfiguration Authentication Confirm frame fails. It *does* say that if the Enrollee fails to decrypt the wrapped data with `k_e`, or if the status code is not `STATUS_OK`, it should revert back to sending Reconfiguration Announcement frames, which I just used as the catch-all failure mode handler, here.


src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -853,6 +857,154 @@ bool ec_ctrl_configurator_t::handle_recfg_announcement(ec_frame_t *encap_frame,
...
+    auto [auth_confirm_frame, auth_confirm_frame_len] = create_recfg_auth_confirm(enrollee_mac, dpp_status, static_cast<uint8_t>(trans_id_attr->data[0]));
+    if (auth_confirm_frame == nullptr || auth_confirm_frame_len == 0) {
+        em_printfout("Failed to create Reconfiguration Authentication Confirm frame");
+        if (wrapped_data) free(wrapped_data);
+        return false;
+    }
+
+    // The Configurator sets a timer for 2 seconds and waits for the Enrollee to start the DPP Configuration (2s dwell)
+    bool sent = m_send_action_frame(sa, reinterpret_cast<uint8_t*>(auth_confirm_frame), auth_confirm_frame_len, 0, 2);
+    if (wrapped_data) free(wrapped_data);
```
tuckerpo-cl, line 1004: self-review: need to `free(auth_confirm_frame)` here 🤦

bcarlson-dev (approved):

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -363,6 +356,84 @@ bool ec_enrollee_t::handle_recfg_auth_request(ec_frame_t *frame, size_t len, uin
...
+        em_printfout("Reconfiguration Authentication Confirm frame is nullptr");
+        return false;
+    }
+
+    size_t attrs_len = len - EC_FRAME_BASE_SIZE;
+
+    auto status_attr = ec_util::get_attrib(frame->attributes, attrs_len, ec_attrib_id_dpp_status);
+    ASSERT_OPT_HAS_VALUE(status_attr, false, "%s:%d: No DPP Status attribute found in Reconfiguration Authentication Confirm frame\n", __func__, __LINE__);
+
+    const auto restart_recfg_announcement = [this]() -> void {
```
bcarlson-dev, line 371: Can we do `free_ephemeral_context` here? That way we start with a fresh slate

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -363,6 +356,84 @@ bool ec_enrollee_t::handle_recfg_auth_request(ec_frame_t *frame, size_t len, uin
...
+
+    if ((memcmp(c_nonce_attr->data, m_eph_ctx().c_nonce, m_c_ctx.nonce_len) != 0) || (memcmp(e_nonce_attr->data, m_eph_ctx().e_nonce, m_c_ctx.nonce_len) != 0)) {
+        em_printfout("Mismatched nonce(s) found in Reconfiguration Authentication Confirm frame, restarting Reconfiguration Announcement frames");
+        free(unwrapped_data);
+        restart_recfg_announcement();
+        return false;
+    }
+
+    auto version_attr = ec_util::get_attrib(unwrapped_data, unwrapped_len, ec_attrib_id_proto_version);
+    ASSERT_OPT_HAS_VALUE_FREE(version_attr, false, unwrapped_data, "%s:%d: No protocol version atteribute found in unwrapped data of Reconfiguration Authentication Confirm frame\n", __func__, __LINE__);
```
bcarlson-dev, line 413: Nit: typo
```suggestion
    ASSERT_OPT_HAS_VALUE_FREE(version_attr, false, unwrapped_data, "%s:%d: No protocol version attribute found in unwrapped data of Reconfiguration Authentication Confirm frame\n", __func__, __LINE__);
```

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -363,6 +356,84 @@ bool ec_enrollee_t::handle_recfg_auth_request(ec_frame_t *frame, size_t len, uin
...
+
+    if (static_cast<uint8_t>(version_attr->data[0]) < 2) {
+        em_printfout("DPP Versio '%d' not supported for Reconfiguration", static_cast<uint8_t>(version_attr->data[0]));
```
bcarlson-dev, line 416: Nit: Typo
```suggestion
        em_printfout("DPP Version '%d' not supported for Reconfiguration", static_cast<uint8_t>(version_attr->data[0]));
```

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -363,6 +356,84 @@ bool ec_enrollee_t::handle_recfg_auth_request(ec_frame_t *frame, size_t len, uin
...
+        free(unwrapped_data);
+        restart_recfg_announcement();
+        return false;
+    }
+
+    // If all's well, we can create and send a Configuration request
+    auto [config_request_frame, config_request_frame_len] = create_config_request();
```
bcarlson-dev, line 423: We still have to use the reconfiguration flags (`EC 8.1.1.22`). Maybe they can be passed as a parameter to `create_config_request` where `CONFIG_REPLACEKEY` does what it would do now (same as regular auth), and `CONFIG_REUSEKEY` just uses the same nak from the current connector (that's what I think it means)?

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -853,6 +857,154 @@ bool ec_ctrl_configurator_t::handle_recfg_announcement(ec_frame_t *encap_frame,
...
+
+
+    auto conn_ctx = get_conn_ctx(enrollee_mac);
+    ASSERT_NOT_NULL(conn_ctx, false, "%s:%d: No known connection context for Enrollee '" MACSTRFMT "'\n", __func__, __LINE__, MAC2STR(sa));
+    auto e_ctx = get_eph_ctx(enrollee_mac);
+    ASSERT_NOT_NULL(e_ctx, false, "%s:%d: Ephemeral context not found for enrollee MAC %s\n", __func__, __LINE__, enrollee_mac.c_str());
+
+    size_t attrs_len = len - EC_FRAME_BASE_SIZE;
+
+    auto trans_id_attr = ec_util::get_attrib(frame->attributes, attrs_len, ec_attrib_id_trans_id);
```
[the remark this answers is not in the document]
tuckerpo-cl: Eph context already has a trans_id field thanks to your foresight months ago, I'll just use that :^)

src/em/prov/easyconnect/ec_enrollee.cpp
tuckerpo-cl: Yeah honestly I'm a bit confused on what the EC spec wants us to do here. The reconfig flags are meant to be a function of the Conn Status object sent from the Enrollee to Configurator during the Recfg Auth Resp step, but as I said in my comments in this PR, spec makes no mention of how to parse the conn status result and map that to reconfig flags.

I'm also equally confused as to what the spec means in terms of the reconfig flags themselves.
`CONFIG_REPLACEKEY` states `The enrollee shall discard the original public/private key and instead the Connector shall use the newly negotiated protocol key pair` but discarding the original pub/priv key is already forced during the Enrollee's handling of the Recfg Auth Request

`CONFIG_REUSEKEY` if I understand it correctly means just the same `P_r, p_r` that was used in the initial Configuration, but again, 6.5.4 states: `if the Enrollee successfully verifies both the C-Connector and the netRole, it generates a protocol keypair (P R , pR )`  as part of the Reconfiguration flow

Overall honestly confused as to how to navigate the whole reconfig flags deal.

I'll address the rest of this PR's comments but I need to give this more thought and/or find a reference codebase that implements Reconfiguration and see what they do, because IMO, the spec really drops the ball here in terms of clarity.

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -1117,8 +1269,6 @@ std::pair<uint8_t *, size_t> ec_ctrl_configurator_t::create_recfg_auth_confirm(s
     uint8_t* attribs = NULL;
     size_t attribs_len = 0;
 
-    // TODO: Add transaction ID outside this function
-    uint8_t trans_id = 0;
```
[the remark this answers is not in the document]
tuckerpo-cl: Actually a new C-Nonce is created (and stored in `e_ctx->c_nonce` as part of `create_recfg_auth_request` on the Configurator, so it should be placing `c_nonce` into the attributes. Good catch.

src/em/prov/easyconnect/ec_enrollee.cpp
bcarlson-dev: I see what you mean now. The real question (with your first part) is, what information from the conn_status object (that the enrollee sent) could possibly tell the configurator whether to reuse or replace the keys. My guess is that they left it open ended because they expected more obvious flags there but instead there's only one flag there which really doesn't make sense on it's own. 

I think we can just keep it hardcoded as REUSE or REPLACE given there's not that much info there. Maybe it's supposed to be the current connection status of the 802.11 network (`STATUS_OK` if it's associated which can happen when it just needs to be reconfigured) and if the status != OK then you should probably just replace the key since they need to re-associate? Who knows.

Regarding your other two questions what I think we can do is:

`CONFIG_REPLACEKEY`: We're already doing this. If we keep doing the config stuff (as you are triggering after this), it will use the protocol keys we just generated through re-config to create the connector then save that one. That is what REPLACE_KEY sounds like to me. Nothing new there.

`CONFIG_REUSEKEY`: We just don't change the `net_access_key`. We can change other parts of the connector if we want we just don't change that when we are generating the config object. 

I think that should work.

REPLACE: Just do what we're doing
REUSE: Don't change any connector keys

src/em/prov/easyconnect/ec_ctrl_configurator.cpp
```diff
@@ -853,6 +857,165 @@ bool ec_ctrl_configurator_t::handle_recfg_announcement(ec_frame_t *encap_frame,
...
+    ASSERT_NOT_NULL(sum.get(), false, "%s:%d: Failed to allocate EC_POINT\n", __func__, __LINE__);
+    if (!EC_POINT_add(conn_ctx->group, sum.get(), C_R.get(), e_ctx->public_resp_proto_key, conn_ctx->bn_ctx)) {
+        em_printfout("Failed to compute C_R + P_R");
+        return false;
+    }
+
+    // sum * c_I = M.x
+    BN_free(e_ctx->m);
+    e_ctx->m = BN_new();
+    ASSERT_NOT_NULL(e_ctx->m, false, "%s:%d: Could not allocate for shared secret M\n", __func__, __LINE__);
```
bcarlson-dev, line 946: You don't need to reallocate, the `compute_ec_ss_x` allocates a new BN

bcarlson-dev (changes requested):

bcarlson-dev (approved):

PR #381, opened 2025-06-13
Fyne cli

soumyasmunshi:


bcarlson-dev:
> Does this have feature parity with the current CLI? I don't see it but I could be missing something. Given it has it's own Makefile and source directory, would it make more sense to keep the current CLI around until this new one has feature parity with it?

Regarding this, I only see three or four views added so I don't think feature parity is there. Can we make this feature complete before removing the current CLIs `main.go` file?

mhughesacn:
Hi Soumya:
1. On the PNG files, please provide the source links so I can look at the corresponding licenses.
3. To be quite clear, please confirm that bundled.go doesn't contain any content that isn't already in the repo e.g. library inclusions.
It doesn't matter if the PNG files are plain or encoded into the bundle.  The key point is the terms of usage, back to point 1.
Thank you,
Martin

PR #383, opened 2025-06-17
Update the fyne cli code for compilation issue

prashant-singh1:
Updated the code to fix the compilation issue, also renamed the certificate with test_ prefix, and corrected the formatting.


bcarlson-dev:
I don't think fyne CLI should be merged since it does not have feature parity with the existing CLI which is being removed.
https://github.com/rdkcentral/unified-wifi-mesh/pull/381#issuecomment-2981120778

Additionally, merging this PR will cause the main branch checks to fail since the underlying issue is that the GitHub action is not being updated for the new fyne CLI makefile. I don't know why the checks are not running for this PR but that will be the result if they are merged

PR #384, opened 2025-06-17
Refactor EasyConnect callbacks into an ec_ops_t struct

tuckerpo-cl:
Also adds send direct encap callback to enrollee

### DIFF
--- a/src/em/prov/easyconnect/ec_configurator.cpp
+++ b/src/em/prov/easyconnect/ec_configurator.cpp
@@ -3,25 +3,17 @@
 #include "ec_util.h"
 #include "util.h"
 
-ec_configurator_t::ec_configurator_t(
-    std::string mac_addr,
-    send_chirp_func send_chirp_notification,
-    send_encap_dpp_func send_prox_encap_dpp_msg,
-    send_dir_encap_dpp_func send_direct_encap_dpp_msg,
-    send_act_frame_func send_action_frame,
-    get_backhaul_sta_info_func backhaul_sta_info_func,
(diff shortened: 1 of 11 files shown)


src/em/prov/easyconnect/ec_configurator.cpp
```diff
@@ -3,25 +3,17 @@
...
+    : m_mac_addr(mac_addr)
 {
+    m_send_chirp_notification    = ops.send_chirp;
+    m_send_prox_encap_dpp_msg    = ops.send_encap_dpp;
+    m_send_dir_encap_dpp_msg     = ops.send_dir_encap_dpp;
+    m_send_action_frame          = ops.send_act_frame;
+    m_get_backhaul_sta_info      = ops.get_backhaul_sta_info;
+    m_get_1905_info              = ops.get_1905_info;
+    m_get_fbss_info              = ops.get_fbss_info;
+    m_can_onboard_additional_aps = ops.can_onboard_additional_aps;
```
bcarlson-dev, line 16: Can you add some NULL checks in the initializers?

bcarlson-dev (changes requested):
🙌

src/em/prov/easyconnect/ec_configurator.cpp
tuckerpo-cl: And take what action if a callback is null? I suppose in `ec_ops.h` we could have dummy callbacks as a default assignment (instead of the current `nullptr`) so that we don't SIGSEGV

PR #394, opened 2025-07-10
Integrated SSID name and passphrase update based on haulType

prashant-singh1:
Reason for change: 1) Integrated SSID name and passphrase update based on haulType
2) Work around for client connection tab crash while STA entries present in db. 
Test Procedure: Observed Fyne cli is working as expected, able to ee the changes ssid and passphrase. 
Risks: Medium
Priority: P1

### DIFF
--- a/src/em/em_net_node.cpp
+++ b/src/em/em_net_node.cpp
@@ -442,6 +442,7 @@ void *em_net_node_t::network_tree_to_json(em_network_node_t *root)
 int em_net_node_t::get_network_tree_node(cJSON *obj, em_network_node_t *root, unsigned int *node_display_ctr)
 {
     cJSON *child_obj, *tmp_obj;
+    int sz = 0;
 
     if (obj->string != NULL) {
         strncpy(root->key, obj->string, strlen(obj->string) + 1);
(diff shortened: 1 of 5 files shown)


soumyasmunshi (approved):

src/em/em_net_node.cpp
```diff
@@ -478,8 +479,11 @@ int em_net_node_t::get_network_tree_node(cJSON *obj, em_network_node_t *root, un
     tmp_obj = child_obj;
 
     while (tmp_obj != NULL) {
-        root->child[root->num_children] = static_cast<em_network_node_t *>(malloc(sizeof(em_network_node_t)));
-        memset(root->child[root->num_children], 0, sizeof(em_network_node_t));
+        sz = sizeof(em_network_node_t) *  2;
```
amarnathhullur, line 482: Why is the size of 2 required? If this is a workaround, mention that in the comment as TBD with details as this needs to be revisited

src/fynecli/BaseTab.go
```diff
@@ -82,6 +82,10 @@ func (b *BaseTab) children(id widget.TreeNodeID) []widget.TreeNodeID {
 
     if nodeType == C.em_network_node_data_type_obj {
         for i := 0; i < int(tree.num_children); i++ {
+            if i >= len(tree.child) {
+                //log.Printf("Child index out of bounds for id %s", id)
+                break
```
amarnathhullur, line 87: is this an error when the index is out of bounds?

src/em/em_net_node.cpp
prashant-singh1: Yes, this is a workaround, will revisit and have proper changes later, Added the comment in code as discussed.

PR #412, opened 2025-07-25
RDKB-61233: Easymesh - WiFi 7 Integration

rakhilpe:
RDKB-61233: Easymesh - WiFi 7 Integration

Reason for change: Add support for WiFi 7 SLO/MLO in Easymesh.
Test Procedure: Ensure basic easymesh configuration and client connectivity works for wifi7 clients.
Risks: Medium
Priority: P1

Signed-off-by: Rakhil P E <rakhilpe001@gmail.com>

### DIFF
--- a/src/dm/dm_easy_mesh.cpp
+++ b/src/dm/dm_easy_mesh.cpp
@@ -88,6 +88,11 @@ dm_easy_mesh_t& dm_easy_mesh_t::operator = (dm_easy_mesh_t const& obj)
         m_policy[i] = obj.m_policy[i];
     }
 
+    m_num_ap_mld = obj.m_num_ap_mld;
+    for (unsigned int i = 0; i < EM_MAX_AP_MLD; i++) {
+        m_ap_mld[i] = obj.m_ap_mld[i];
+    }
+
     sta = static_cast<dm_sta_t *> (hash_map_get_first(obj.m_sta_map));
     while (sta != NULL) {
         dm_easy_mesh_t::macbytes_to_string(sta->m_sta_info.id, sta_mac_str);
@@ -184,6 +189,17 @@ int dm_easy_mesh_t::commit_config(dm_easy_mesh_t& dm, em_commit_target_t target)
 				}
 			}
 		}
+
+		em_printfout("Number of AP MLDs to commit : %d", dm.m_num_ap_mld);
+		for (i = 0; i < dm.m_num_ap_mld; i++) {
+			em_ap_mld_info_t *src_mld_info = dm.m_ap_mld[i].get_ap_mld_info();
+			if (src_mld_info->num_affiliated_ap > 0) {
+				update_ap_mld_info(src_mld_info);
+			} else {
+				em_printfout("Skipping MLD[%d] as no affiliated APs found", i);
+			}
+		}
+
 	}
     return false;
 }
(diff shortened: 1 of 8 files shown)


src/dm/dm_easy_mesh.cpp
```diff
@@ -2778,6 +2794,108 @@ void dm_easy_mesh_t::update_scan_results(em_scan_result_t *scan_result)
...
 
+em_ap_mld_info_t *dm_easy_mesh_t::get_ap_mld_frm_bssid(mac_address_t bss_id)
+{
+    unsigned int i, j;
+    em_ap_mld_info_t *ap_mld_info = NULL;
+
+    for (i = 0; i < m_num_ap_mld; i++) {
+        ap_mld_info = &m_ap_mld[i].m_ap_mld_info;
+        for (j = 0; j < ap_mld_info->num_affiliated_ap; ++j) {
+            if (memcmp(ap_mld_info->affiliated_ap[j].mac_addr, bss_id, sizeof(mac_address_t)) == 0) {
```
amarnathhullur, line 2805: if this does not match for any then since in 2803 ap_mld_info is set to a non-NULL value it will send the last assigned value. You should probably have a different temp value to check the mac_addr and assign the temp pointer to ap_mld_info if it matches.

src/dm/dm_easy_mesh.cpp
```diff
@@ -2778,6 +2794,108 @@ void dm_easy_mesh_t::update_scan_results(em_scan_result_t *scan_result)
...
+    target_mld->str = ap_mld_info->str;
+    target_mld->nstr = ap_mld_info->nstr;
+    target_mld->emlsr = ap_mld_info->emlsr;
+    target_mld->emlmr = ap_mld_info->emlmr;
+
+    // Loop through all affiliated APs
+    for (int j = 0; j < ap_mld_info->num_affiliated_ap; j++) {
+        em_affiliated_ap_info_t *input_ap = &ap_mld_info->affiliated_ap[j];
+        em_affiliated_ap_info_t *target_aff_ap = NULL;
+        bool aff_ap_found = false;
```
amarnathhullur, line 2855: Need more understanding of this. This declaration shouldn't it be outside of the int j=0 loop?

src/dm/dm_easy_mesh.cpp
```diff
@@ -184,6 +189,17 @@ int dm_easy_mesh_t::commit_config(dm_easy_mesh_t& dm, em_commit_target_t target)
...
 			}
 		}
+
+		printf("%s:%d Number of AP MLD to commit %d\n", __func__, __LINE__, dm.m_num_ap_mld);
+		for (i = 0; i < dm.m_num_ap_mld; i++) {
+			em_ap_mld_info_t *src_mld_info = dm.m_ap_mld[i].get_ap_mld_info();
+			if (src_mld_info->num_affiliated_ap > 0) {
+				update_ap_mld_info(src_mld_info);
+			} else {
+				printf("%s:%d Skipping MLD[%d] as no affiliated APs found\n", __func__, __LINE__, i);
```
amarnathhullur, line 199: General comment use em_printfout instead of printf for all the printf calls you are adding.
em_printfout adds a func and line so no need of func and line as arguments and also \n at the end can be removed.

src/dm/dm_easy_mesh.cpp
amarnathhullur, line 1: General comment use em_printfout instead of printf for all the printf calls you are adding.
em_printfout adds a func and line so no need of func and line as arguments and also \n at the end can be removed.

amarnathhullur (changes requested):

src/em/config/em_configuration.cpp
amarnathhullur, line 1: General comment use em_printfout instead of printf for all the printf calls you are adding.
em_printfout adds a func and line so no need of func and line as arguments and also \n at the end can be removed.

src/em/config/em_configuration.cpp
```diff
@@ -1639,6 +1638,25 @@ int em_configuration_t::handle_topology_response(unsigned char *buff, unsigned i
...
+            continue;
+
+        } else {
+            found_ap_mld = true;
+            break;
+       }
+    }
+
+    if (found_ap_mld == true) {
+        printf("%s:%d: Found AP MLD details in topology response mesaage\n", __func__, __LINE__);
```
amarnathhullur, line 1656: typo in the string

amarnathhullur (changes requested):

amarnathhullur (approved):

PR #415, opened 2025-07-31
RDKBACCL-578: EasyMesh Agent integration in BPI Extender device

ap934:
Reason for change: Added EM_EXTENDER flag for bpi extender builds
Test procedure: Build should go through, and only agent and al-sap should get compiled in extender
Risks: None

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -16,5 +16,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ##########################################################################
+if EM_EXTENDER
+SUBDIRS = al-sap agent
+else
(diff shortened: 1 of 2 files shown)


src/Makefile.am
```diff
@@ -16,5 +16,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ##########################################################################
+if EM_EXTENDER
+SUBDIRS = agent al-sap
```
kpandu987, line 20: this line will cause error in build time..
Please check once

ap934: Okay @kpandu987 
Will update.

PR #423, opened 2025-08-07
Implements BSS Configuration Response, Result, and Agent List

bcarlson-dev:
Implements:
- Handle BSS Configuration Request message
- Create / Handle BSS Configuration Response message
- Create / Handle BSS Configuration Result message
- Create / Handle Agent List message

Other changes:
- Moves implementations for these to `em_configuration*` to re-use existing TLV handlers and creation methods
- Moves EasyMesh DPP callbacks to `em*` for use when creating  BSS Configuration Response messages.
- Adds support for deliminators in `macstr_to_vector`
- Adds iterator functions for tlvs

This currently exists in a vacuum, with no code sending BSS Configuration Requests and therefore BSS Configuration Responses, Results, and Agent Lists not being sent but that should soon change

Regular Ethernet WSC onboarding has been tested with this code

### DIFF
--- a/src/em/config/em_configuration.cpp
+++ b/src/em/config/em_configuration.cpp
@@ -46,6 +46,8 @@
 #include "em_crypto.h"
 #include "em.h"
 #include "em_cmd_exec.h"
+#include "cjson_util.h"
+#include "ec_ctrl_configurator.h"
 
 // Initialize the static member variables
 unsigned short em_configuration_t::msg_id = 0;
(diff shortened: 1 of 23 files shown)


src/em/config/em_configuration.cpp
```diff
@@ -3697,7 +4530,19 @@ void em_configuration_t::process_msg(unsigned char *data, unsigned int len)
...
+            break;
+        case em_msg_type_bss_config_rsp:
+            handle_bss_config_rsp_msg(data, len, src_al_mac);
+            break;
+        case em_msg_type_bss_config_res:
+            handle_bss_config_res_msg(data, len, src_al_mac);
+            break;
+        case em_msg_type_agent_list:
+            handle_agent_list_msg(data, len, src_al_mac);
+            break;
```
tuckerpo-cl, line 4545: Not strictly necessary but might want to check `get_service_type()` before passing off to handlers in case in some time in the future addressing becomes broken again and Controller receives a BSS Config Resp Msg or Agent receives a BSS Config Req :shrug:

src/em/prov/em_provisioning.cpp
```diff
@@ -625,7 +382,7 @@ int em_provisioning_t::handle_1905_rekey_msg(uint8_t *buff, unsigned int len){
     return 0;
 }
 
-int em_provisioning_t::handle_1905_encap_eapol_msg(uint8_t *buff, unsigned int len){
+int em_provisioning_t::handle_1905_encap_eapol_msg(uint8_t *buff, unsigned int len, uint8_t src_al_mac[ETH_ALEN]){
```
tuckerpo-cl, line 385: Seems `src_al_mac` is unused here, maybe `(void)`, `[[maybe_unused]]`, or remove altogether (or perhaps it's meant to replace the `uint8_t *src_mac = hdr->src` local a bit futher down?)

tuckerpo-cl (changes requested):
Mostly doc nits but maybe a bug or two

src/em/config/em_configuration.cpp
bcarlson-dev: Good point

PR #425, opened 2025-08-12
Topology update and Crash fixes

amarnathhullur:


src/ctrl/em_network_topo.cpp
```diff
@@ -88,6 +94,90 @@ em_network_topo_t *em_network_topo_t::find_topology_by_bh_associated(mac_address
...
+	em_network_topo_t *topo;
+
+	dm_easy_mesh_t::macbytes_to_string(bss_mac, bss_mac_str);
+	for (i = 0; i < m_data_model->m_num_bss; i++) {
+		if (memcmp(m_data_model->m_bss[i].m_bss_info.id.bssid, bss_mac, sizeof(mac_address_t)) == 0) {
+			printf("%s:%d: Found topology of bss mac: %s num_child_topo:%d\n", __func__, __LINE__,
+				   bss_mac_str, this->m_num_topologies);
+			return this;
+		}
+	}
```
bcarlson-dev, line 150: I think this can be replaced by:

```suggestion
	em_bss_info_t* bss_info = m_data_model->get_bss_info_with_mac(bss_mac);
    EM_ASSERT_NULL(bss_info, this, "Found topology of bss mac: %s num_child_topo:%d", bss_mac_str, this->m_num_topologies);
```

`EM_ASSERT_NULL` will automatically expand and check if `bss_info` is NULL and if it is (meaning in this case it was found) then it will return `this`. The `__func__` and `__LINE__` are automatically used by the `em_printfout` macro which is called in the `EM_ASSERT_NULL`.

It's equivalent to:
```
if (bss_info == NULL) {
   em_printfout("Found topology of bss mac: %s num_child_topo:%d", bss_mac_str, this->m_num_topologies);
   return this;
}

File: src/dm/dm_bss.cpp [hunk 2]
```diff
@@ -211,6 +211,7 @@ void dm_bss_t::encode(cJSON *obj, bool summary)
     cJSON_AddStringToObject(obj, "SSID", m_bss_info.ssid);
     cJSON_AddBoolToObject(obj, "Enabled", m_bss_info.enabled);
     cJSON_AddStringToObject(obj, "TimeStamp", m_bss_info.timestamp);
+    cJSON_AddNumberToObject(obj, "VapMode", m_bss_info.vap_mode);
```
bcarlson-dev, line 214: I intentionally left this part out when I added vap_mode (and most importantly in the OneWifi `webconfig_easymesh_translator` so that easymesh couldn't change the vap mode on it's own which could cause undefined behavior. By leaving it out of the EM encode it was absolutely clear that it could not be changed from the EM side. If you think it's best to add it then that's perfectly fine by me, I just wanted to explain my rationale

amarnathhullur: I need this as I have to differentiate between sta and ap in controller which is not possible just by haul type alone.

src/agent/em_agent.cpp
```diff
@@ -1648,7 +1648,9 @@ em_t *em_agent_t::find_em_for_msg_type(unsigned char *data, unsigned int len, em
             em = al_em;
             break;
         default:
-            printf("%s:%d: Frame: %d not handled in agent\n", __func__, __LINE__, htons(cmdu->type));
+            if (htons(cmdu->type) != em_msg_type_topo_disc) {
```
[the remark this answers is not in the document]
amarnathhullur: Done

src/ctrl/em_ctrl.cpp
```diff
@@ -796,7 +797,9 @@ em_t *em_ctrl_t::find_em_for_msg_type(unsigned char *data, unsigned int len, em_
 	        break;
 
         default:
-            printf("%s:%d: Frame: 0x%04x not handled in controller\n", __func__, __LINE__, htons(cmdu->type));
+            if (htons(cmdu->type) != em_msg_type_topo_disc) {
```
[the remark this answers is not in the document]
amarnathhullur: Done

PR #431, opened 2025-08-26
Initial changes for network topology graph tab

prashant-singh1:


prashant-singh1:
> Hi Prashant: Where did the new images in this pull req come from? (That includes icons and images of iPhone, iPad etc.)

I have downloaded the below icons from the links shared.
Ipad:
https://share.google/images/AgpJ98zolbq6XLniq

android:
https://share.google/images/iJhwdKBOsF5OaYDnt

iphone:
https://share.google/images/l8AfTxAoPJ4KRx6Ga

few icons have been shared by @soumyasmunshi , Will check with him on the source

mhughesacn (changes requested):
Thanks Prashant.  These three images cannot be used because there are associated licenses with conditions.  To be honest, the easiest way to get images of iPhone, iPad and Android phone would be to take your own photographs which would come without any external license obligations.  Otherwise, you need to find a suitable image site - pexels and pixabay licenses were allowed by Comcast Legal in the past, also some of the Creative Commons licenses are good.  But you will have to store in separate directories with more license info.

The images from Android and iPhone/iPad also contain Google or Apple content, but I don't think this is worth pursuing especially if it's a photo of someone's phone ("Fair Use").

Also, where does topology.go come from?  All significant code files should have headers unless autogenerated.

Still interested in the other icons as well.

Thank you!

amarnathhullur:
This change is part of fynecli which is deprecated now. Thus closing this PR

PR #432, opened 2025-08-27
Changes for Radiolist DB population, Sta encode for topology, Station Client Type incorrect

amarnathhullur:


src/em/metrics/em_metrics.cpp
```diff
@@ -216,7 +216,7 @@ int em_metrics_t::handle_associated_sta_link_metrics_resp(unsigned char *buff, u
 
     while ((tlv->type != em_tlv_type_eom) && (tmp_len > 0)) {
         if (tlv->type == em_tlv_type_vendor_sta_metrics) {
-            handle_assoc_sta_vendor_link_metrics_tlv(tlv->value);
+            handle_assoc_sta_vendor_link_metrics_tlv(tlv->value, htons(tlv->len));
```
bcarlson-dev, line 219: So I understand that this is what is reflected throughout the code but I've been confused by the use of `htons` in the codebase for both sending and recieving. For sending it makes sense since it's changing from the endianness of the machine to big-endian (network) but on the recieving end, you're recieving a big-endian 16 bit unsigned int. Then, converting it to big endian based on the endian-ness of your current machine (which if you're on a little-endian machine would mean swapping them but if you're on a big-endian machine would mean keeping them the same). Why not just use `ntohs` on the recieving end since that converts from the network order (big-endian) to the host-order (whatever that may be).

I'm not saying this must be changed since it's a pattern in the code and it's still following the spec since it's being sent in big-endian but I was just wondering if you could shed some light on it.

src/em/metrics/em_metrics.cpp
```diff
@@ -463,7 +463,7 @@ int em_metrics_t::handle_ap_metrics_response(unsigned char *buff, unsigned int l
...
```
bcarlson-dev, line 466: Same here

bcarlson-dev (commented):
Looks good to me but if you could provide a bit of insight on those small things that would be great

src/em/metrics/em_metrics.cpp
amarnathhullur: Agree with you, made the change to use ntohs instead of htons

src/em/metrics/em_metrics.cpp
amarnathhullur: Agree with you, made the change to use ntohs instead of htons

bcarlson-dev (approved):

PR #434, opened 2025-09-02
Enabling AL_SAP by default in OpenWRT and updating current class and channel in network topology

amarnathhullur:
Two commits added as part of this PR
1) In OpenWRT makefile enabling AL_SAP by default to use ieee1905 rust.
2) Add current operating class and channel in network topology encode

### DIFF
--- a/src/ctrl/em_network_topo.cpp
+++ b/src/ctrl/em_network_topo.cpp
@@ -37,6 +37,11 @@ void em_network_topo_t::encode(cJSON *parent)
 	cJSON *dev_obj, *child_obj, *radio_list_obj, *radio_obj, *bss_list_obj, *bss_obj, *bh_obj, *sta_list_obj;
 	unsigned int i, j;
 
+	if (m_data_model == NULL) {
+		em_printfout("Data model is NULL, cannot encode topology");
+		return;
+	}
+
 	dev_obj = cJSON_AddObjectToObject(parent, "Device");
 	m_data_model->m_device.encode(dev_obj, true);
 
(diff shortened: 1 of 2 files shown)


src/ctrl/em_network_topo.cpp
```diff
@@ -44,6 +44,20 @@ void em_network_topo_t::encode(cJSON *parent)
 	for (i = 0; i < m_data_model->m_num_radios; i++) {
 		radio_obj = cJSON_CreateObject();
 		m_data_model->m_radio[i].encode(radio_obj, em_get_radio_list_reason_radio_summary);
+		dm_easy_mesh_t *dm = m_data_model;
+		if (dm != NULL) {
```
carlsondev, line 48: I think the `m_data_model != NULL` check would be better positioned at the beginning of the method with an early return since it's a global variable and that way the check is not performed every iteration of the loop.
```cpp
if (dm == NULL) {
    return;
}
```
(around line 39)

src/ctrl/em_network_topo.cpp
```diff
@@ -44,6 +44,20 @@ void em_network_topo_t::encode(cJSON *parent)
 	for (i = 0; i < m_data_model->m_num_radios; i++) {
 		radio_obj = cJSON_CreateObject();
 		m_data_model->m_radio[i].encode(radio_obj, em_get_radio_list_reason_radio_summary);
+		dm_easy_mesh_t *dm = m_data_model;
```
carlsondev, line 47: I see that you're using this assignment as a bit of a shorthand for `m_data_model`. That's fine except that `m_data_model` is used later in these changes (as well as later in this method) so it essentially is only used for two lines. I think that this should either be removed with the following two lines using `m_data_model` instead of `dm` or it should be move outside of the loop and all of the uses of `m_data_model` (both in this loop and the following one should be changed to use it).

src/ctrl/em_network_topo.cpp
amarnathhullur: Addressed in the latest commit

src/ctrl/em_network_topo.cpp
amarnathhullur: Addressed in the latest commit

carlsondev (approved):

PR #435, opened 2025-09-03
Default Vlan ID for CLI

Nikita-Hakai:
Default Vlan ID for CLI to be used in Network Topology


src/dm/dm_bss.cpp
amarnathhullur, line 1: Could you put a comment that the hardcoding of vlan id is temporary and should be updated when the traffic separation changes are integrated from OneWifi and Agent

src/dm/dm_bss.cpp
```diff
@@ -178,29 +178,35 @@ void dm_bss_t::encode(cJSON *obj, bool summary)
     mac_addr_str_t  mac_str;
     unsigned short i;
 	em_short_string_t	haul_type_str;
+    unsigned int vlan = 0;
```
amarnathhullur, line 181: would be good to change the name as vlan_id

amarnathhullur (approved):

PR #443, opened 2025-09-10
Update AP MLD and STA MLD Mac Address details in CLI

rakhilpe:


src/dm/dm_sta.cpp
```diff
@@ -232,6 +237,7 @@ void dm_sta_t::encode(cJSON *obj, em_get_sta_list_reason_t reason)
         encode_beacon_report(obj);
     } else if (reason == em_get_sta_list_reason_topology) {
         cJSON_AddStringToObject(obj, "SSID", m_sta_info.ssid);
+        cJSON_AddStringToObject(obj, "MLDAddr", m_sta_info.multi_link);
```
amarnathhullur, line 240: Add MLDAddr only if multi_link is present or non-zero

src/dm/dm_sta.cpp
```diff
@@ -455,6 +465,28 @@ void dm_sta_t::decode_sta_capability(dm_sta_t *sta)
...
+                    ext_ptr = tag->value + 1;
+                    ext_len = tag->length - 1;
+
+                    // Minimum required length: 2 (multilink_control) + 1 (common_info_len) + 6 (MLD MAC)
+                    if (ext_len >= 9) {
+                        ext_ptr += 2; // Skip multilink_control
+                        common_info_len = ext_ptr[0];
+                        ext_ptr++;
+
+                        if (common_info_len >= 6) {
```
amarnathhullur, line 479: Change 6 to have the corresponding #define for MACaddress length

src/dm/dm_sta.cpp
```diff
@@ -455,6 +465,28 @@ void dm_sta_t::decode_sta_capability(dm_sta_t *sta)
...
+                            snprintf(sta->m_sta_info.multi_link, sizeof(sta->m_sta_info.multi_link),
```
amarnathhullur, line 480: Use util::mac_to_string instead of printf to copy the mac address to multi_link

PR #444, opened 2025-09-10
Changed DPP URI handling from JSON to standard URI

bcarlson-dev:
Note: There still has to be some use of JSON because the "CLI" uses JSON when sending data but the actual URI is saved and read as the normal format. 

For example. 
```bash
root@OpenWrt:~# cat /nvram/DPPURI.txt
DPP:V:2;M:a22efd60e5c9;C:81/12;K:MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE1oFWzqZ0mQF6HQt5ulPzAZUgHM8w75SWpFSxdSmxHsoFJG1Mufln+7BTYUiH4kXghol1hX6hxO5VhfPzTnm8QA==;;
```

### DIFF
--- a/src/cli/utils.go
+++ b/src/cli/utils.go
@@ -122,6 +122,8 @@ func readJSONFile(filePath string) ([]etree.Node, error) {
 	return nodes, nil
 }
 
+
+
 // convertNodesToJSON converts an etree.Node structure back into a generic interface{}
 // that can be marshaled to JSON. It handles objects, arrays, strings, numbers and booleans.
 //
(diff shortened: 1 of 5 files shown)


src/cli/utils.go
```diff
@@ -187,6 +189,55 @@ func convertNodesToJSON(nodes []etree.Node) interface{} {
...
+
+	// Trim whitespace/newlines from the content
+	uriContent := strings.TrimSpace(string(content))
+
+	// Create JSON structure with URI object containing URI key
+	jsonData := map[string]interface{}{
+		"URI": map[string]interface{}{
+			"URI": uriContent,
+		},
+	}
```
tuckerpo-cl, line 230: Comment implies that this intermediate JSON representation would be
```json
{
  "URI": "DPP:C:81/1:some-key-here;"
}
```

but this seems like it'd make it doubly-nested
```json
{
  "URI": {
    "URI": "DPP:C:81/1:some-key-here;"
  }
}
```

Is this intentional?

tuckerpo-cl (approved):
LGTM other than my singular comment which may just be my misunderstanding

PR #452, opened 2025-09-18
STA details in Database is getting duplicated when connecting/disconnecting

sriraaman-c:
Reason for change : The "update" query which is available only handles the MAC of the client, if the same MAC is connected across different bands, then the update query modifies both. Hence extended the query to check BSSID also.


src/db/db_easy_mesh.cpp
```diff
@@ -162,10 +162,17 @@ int db_easy_mesh_t::update_row(db_client_t& db_client, ...)
     format[strlen(format) - 2] = 0;
     snprintf(tmp, sizeof(db_query_t), " where %s = ", m_columns[0].m_name);
     snprintf(format + strlen(format), sizeof(format) - strlen(format), "%s", tmp);
-
     snprintf(format + strlen(format), sizeof(format) - strlen(format), "%s", get_column_format(col_fmt, 0));
     format[strlen(format) - 2] = 0;
 
+    /* Extending query with extra BSSID check only for STAList table updation */
```
Nikita-Hakai, line 168: Actually update_row() in db_easy_mesh.cpp generally contain common db operations, anything specific to a functionality like stalist should be done in the dm_sta_list.cpp file is what I think.

PR #476, opened 2025-10-28
Enrollee: handle DPP_STATUS_CONFIGURE_PENDING and DPP_STATUS_NEW_KEY_NEEDED

tuckerpo-cl:
This improves interop with non-RDK Configurators. `DPP_STATUS_CSR_BAD` is a large cans of worms and will be a different PR.


amarnathhullur:
Old change, needs to be refreshed if required

PR #481, opened 2025-11-04
RDKBACCL-1221: Channel preference report to contain information of all radios.

AbijithNayak:
The channel preference and selection messages were dependent on presence of radio id.
 Modified to message based on the AL mac of the agent and updated TLVs based on IEEE1905 standard.

 Test report: Verified the channel preference and selection messages and TLVs based on ieee1905 formats

### DIFF
--- a/src/ctrl/em_ctrl.cpp
+++ b/src/ctrl/em_ctrl.cpp
@@ -629,6 +629,7 @@ em_t *em_ctrl_t::find_em_for_msg_type(unsigned char *data, unsigned int len, em_
     em_2xlong_string_t key;
     unsigned int i;
     bool found;
+    em_string_t al_mac_str;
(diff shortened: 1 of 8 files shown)


src/ctrl/em_ctrl.cpp
```diff
@@ -704,6 +705,16 @@ em_t *em_ctrl_t::find_em_for_msg_type(unsigned char *data, unsigned int len, em_
 
         case em_msg_type_topo_resp:
         case em_msg_type_channel_pref_rprt:
+            dm_easy_mesh_t::macbytes_to_string(hdr->dst, al_mac_str);
```
amarnathhullur: hdr->dst will be pointing to controller's al_mac.  Are we handling the topo_resp and channel_pref_rprt on the controller's al_mac object?  Not sure if this would work.

PR #482, opened 2025-11-05
RDKBACCL-1193: Parameter getters for controller

meycint:
Reason for change: Network Optimizer needs data, which is collected
  over underlying bus. Parameter getters for required objects over
  bus is missing. These changes provide essential parameters.
Risks: Low


src/ctrl/tr_181/tr_181_param.cpp
```diff
@@ -62,4 +81,2066 @@ bus_error_t em_ctrl_t::get_device_wifi_dataelements_network_controllerid (char *
...
+        rc = raw_data_set(p_data, 0U);
+    } else {
+        printf("Invalid param: %s\n", param);
+        rc = bus_error_invalid_input;
+    }
+
+    return rc;
+}
+
+#if 0
```
amarnathhullur: Why is there #if 0 for this code. If not required, can it be removed?

PR #493, opened 2025-11-14
RDKBDEV-3310: onewifi_em_ctrl is not properly initialized when DB doe…

vlad-safonov:
…sn't exist

Reason for change: when em_ctrl.service is started for the first time when DB doesn't exist newifi_em_ctrl is not initialized properly and additional service restart is required.
This fix detects if database is empty after load_tables() and executes setup script to fill DB tables directly from /usr/bin/onewifi_em_ctrl, after that calls one more time load_tables() like booting with existing DB.

Test procedure:
1) remove DB and relevant mysql db flags from /nvram: mysql_db_account_exists and mysql_db_data_exists; 2) restart em_ctrl.service;
3) sanity check onewifi_em_ctrl: em cli shold work, configuration done properly, em_ctrl.log should contain message about filling DB; 4)  restart em_ctrl.service again (withiout removing DB) and sanity check onewifi_em_ctrl: there should be no regression;

Priority: P2
Risks: Low

### DIFF
--- a/src/ctrl/dm_easy_mesh_ctrl.cpp
+++ b/src/ctrl/dm_easy_mesh_ctrl.cpp
@@ -2069,7 +2069,6 @@ int dm_easy_mesh_ctrl_t::init(const char *data_model_path, em_mgr_t *mgr)
 
        //Load tables and update rc to check it for non-empty database
        rc = load_tables();
-       assert(rc != -1);
     }
 
     if (rc != 0) {


pradeeptakdas:
Internal ticket for tracking https://ccp.sys.comcast.net/browse/RDKCOM-5469

vlad-safonov:
Hi @amarnathhullur, yes correct and I have to prepare that PR also, will do it today

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -2059,7 +2059,20 @@ int dm_easy_mesh_ctrl_t::init(const char *data_model_path, em_mgr_t *mgr)
...
+
+    //Database is empty and need to fill it, then load tables with data again
+    if (rc == -1) {
+       //Assuming this will not fail, and there is known setup script to fill data
+       printf("%s:%d: data base empty ... fill it from /usr/ccsp/EasyMesh/setup_mysql_db_post.sh\n", __func__, __LINE__);
+       std::system("/usr/ccsp/EasyMesh/setup_mysql_db_post.sh");
+
+       //Load tables and update rc to check it for non-empty database
+       rc = load_tables();
+       assert(rc != -1);
```
amarnathhullur: Don't assert here. In other targets the above script if not present then load_tables may fail again. Better not to assert and let the controller continue to run

amarnathhullur (approved):

PR #496, opened 2025-11-20
RDKBACCL-1207: Topology updates to orchestrator

Nikita-Hakai:
- Topology update on a client assoc/disassoc and publish to orchestrator is enabled.
- Method to receive policy and node sync is also enabled.


src/ctrl/tr_181/tr_181_param.cpp
```diff
@@ -39,6 +39,60 @@ bus_error_t curops_tget_params(dm_easy_mesh_t *dm, const char *root, em_radio_in
...
+        em_printfout("Node sync publish fail\n",__func__, __LINE__);
+    }
+
+    return bus_error_success;
+}
+
+bus_error_t em_ctrl_t::policy_config(char *event_name, raw_data_t *p_data, bus_user_data_t *user_data)
+{
+    em_printfout(" Event rcvd: %s\n Policy cfg is: \n%s\n", event_name, p_data->raw_data.bytes);
+
```
amarnathhullur: is there a TBD item here?

src/em/config/em_configuration.cpp
```diff
@@ -5250,6 +5238,7 @@ void em_configuration_t::process_msg(unsigned char *data, unsigned int len)
                     em_radios.clear();
                     // update network topology here
                     get_mgr()->update_network_topology();
+                    dm->m_topo_changed = true;
```
amarnathhullur: Direct access of member variable of a class here is not correct, use a helper member function

src/ctrl/tr_181/tr_181_param.cpp
Nikita-Hakai: > is there a TBD item here?

Node sync and policy are to be done. Right now just publishing dummy data for node sync.

PR #501, opened 2025-11-25
RDKBACCL-1207: Topology updates to orchestrator

Nikita-Hakai:
- Topology update on a client assoc/disassoc and publish to orchestrator is enabled.
- Method to receive policy and node sync is also enabled.


amarnathhullur:
@Nikita-Hakai , I am unable to merge this as it is showing conflict while rebasing. Could you please rebase the change

PR #517, opened 2025-12-05
RDKBACCL-1321: GUI Improvement to handle wifi reset popup and topolog…

prashant-singh1:
Handle the wifi reset confirmation dialog popup in better way to avoid duplicate pop-ups
Modified topology graph to reload on click on tab and every 1 min if we stay on same page.
Also improve some minor Gui issues

### DIFF
--- a/src/rdkb-cli/static/script.js
+++ b/src/rdkb-cli/static/script.js
@@ -366,9 +366,9 @@ async handleWifiResetApply() {
       resetBtn.classList.remove("is-loading");
       resetBtn.removeAttribute("aria-busy");
     }
-      this._blockShortcutsDuringReset?.(false);
-      this._wifiResetInProgress = false;
-    }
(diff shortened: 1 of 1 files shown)


src/rdkb-cli/static/script.js
```diff
@@ -304,6 +309,68 @@ class EasyMeshController {
...
+    // Restore UI state
+    if (resetBtn) {
+      resetBtn.disabled = false;
+      resetBtn.textContent = originalBtnText ?? "Apply Reset";
+      resetBtn.classList.remove("is-loading");
+      resetBtn.removeAttribute("aria-busy");
+    }
+      this._blockShortcutsDuringReset?.(false);
+      this._wifiResetInProgress = false;
+    }
```
amarnathhullur: Indentation of above 3 lines seems incorrect

PR #524, opened 2025-12-10
RDKBACCL-1330: Updated current tx_power in operating channel report TLV

sundram0711:
- Updated TX power in Unified Mesh `em_op_class_info_t` from onewifi.
- Populate TX power in Operating Channel Report TLV within the Operating Channel Report message

Closes: RDKBACCL-1330
Signed-off-by: Sundram Patel <sundram.p@tataelxsi.co.in>

### DIFF
--- a/src/em/channel/em_channel.cpp
+++ b/src/em/channel/em_channel.cpp
@@ -746,6 +746,7 @@ short em_channel_t::create_operating_channel_report_tlv(unsigned char *buff)
 	em_op_class_ch_rprt_t *rprt_channel;
 	dm_op_class_t	*op_class;	
 	unsigned char *tmp;
+	int8_t curr_txpwr_dbm = 0;
 
 	dm = get_data_model();
 
@@ -766,14 +767,14 @@ short em_channel_t::create_operating_channel_report_tlv(unsigned char *buff)
 			len += static_cast<short unsigned int> (sizeof(em_op_class_ch_rprt_t));
 			tmp += sizeof(em_op_class_ch_rprt_t);
 
-			len += static_cast<short unsigned int> (sizeof(unsigned char));
-			tmp += sizeof(unsigned char);	
-
 			rprt_channel = reinterpret_cast<em_op_class_ch_rprt_t *> (tmp);
 
 			rprt_op_class->op_classes_num++;
 		}
 	}	
+	curr_txpwr_dbm = static_cast<int8_t>(op_class->m_op_class_info.tx_power);
+	memcpy(buff + len, &curr_txpwr_dbm, sizeof(curr_txpwr_dbm));
+	len += static_cast<short>(sizeof(curr_txpwr_dbm));
 
 	return len;
 }


src/em/channel/em_channel.cpp
```diff
@@ -763,12 +763,10 @@ short em_channel_t::create_operating_channel_report_tlv(unsigned char *buff)
 			tmp = reinterpret_cast<unsigned char *> (rprt_channel);
 			rprt_channel->op_class = static_cast<unsigned char> (op_class->m_op_class_info.op_class);
 			rprt_channel->channel = static_cast<unsigned char> (op_class->m_op_class_info.channel);
+			rprt_channel->current_transmit_power = (int8_t)op_class->m_op_class_info.tx_power;
 			len += static_cast<short unsigned int> (sizeof(em_op_class_ch_rprt_t));
 			tmp += sizeof(em_op_class_ch_rprt_t);
 
-			len += static_cast<short unsigned int> (sizeof(unsigned char));
-			tmp += sizeof(unsigned char);	
-
```
amarnathhullur: Removing of the above two lines sounds appropriate, but the setting of the transmit_power should be outsiof the loop and then addition of one length of unsigned char to consider transmit_power added should be done.

amarnathhullur (changes requested):

src/em/channel/em_channel.cpp
sundram0711: Done. thanks

inc/em_base.h
```diff
@@ -1046,6 +1046,7 @@ typedef struct {
     mac_address_t  ruid;
     unsigned char  op_classes_num;
     em_op_class_ch_rprt_t  op_classes[0];
+    int8_t  current_transmit_power;
```
amarnathhullur: Adding the current_transmit_power here would also not be correct as the op_classes[0] is a 0 byte structure which would not allocate any memory. So the offset of current_transmit_power would not be correct if there are multiple op_classes included. 
The correct approach would be to add transmit_power separately taking it as a parameter to create_operating_channel_report_tlv and adding it at the end after the op_classes have been added to the buffer.

src/em/channel/em_channel.cpp
```diff
@@ -766,14 +766,13 @@ short em_channel_t::create_operating_channel_report_tlv(unsigned char *buff)
...
 			rprt_channel = reinterpret_cast<em_op_class_ch_rprt_t *> (tmp);
 
 			rprt_op_class->op_classes_num++;
 		}
 	}	
+	rprt_op_class->current_transmit_power = tx_power;
```
amarnathhullur: Where is the tx_power defined or declared? Isn't this change resulting in a compilation error as I don't see declaration of tx_power.

sundram0711: Oops! I made a mistake here; I had to push the changes from another workspace. 
I'll push again.

inc/em_base.h
sundram0711: Done. I believe the implementation is correct now, and since TX power is derived internally, there’s no need to pass it as an argument to the create_operating_channel_report_tlv() function.
If you think any further changes are required, please let me know.

src/em/channel/em_channel.cpp
sundram0711: fixed.

inc/em_base.h
sundram0711: In the latest mesh, I'm not receiving the operating channel report message.

amarnathhullur (changes requested):
Looks good, can you please update the commit text and header with JIRA. Open the JIRA in jira.rdkcentral.com similar to https://jira.rdkcentral.com/jira/browse/RDKBACCL-1340

sundram0711:
> Looks good, can you please update the commit text and header with JIRA. Open the JIRA in jira.rdkcentral.com similar to https://jira.rdkcentral.com/jira/browse/RDKBACCL-1340

Okay sure

sundram0711:
> Looks good, can you please update the commit text and header with JIRA. Open the JIRA in jira.rdkcentral.com similar to https://jira.rdkcentral.com/jira/browse/RDKBACCL-1340

done

amarnathhullur (approved):

PR #525, opened 2025-12-10
RDKBACCL-1329 : TLV Length correction for AP Autoconfiguration M2

SachinMChougale:
1. Connection type flag len changed to 1B
2. Primary Device type len changed to 8B
3. OS Version Len changed to 4B

### DIFF
--- a/src/em/config/em_configuration.cpp
+++ b/src/em/config/em_configuration.cpp
@@ -2024,8 +2024,9 @@ unsigned short em_configuration_t::create_m2_msg(unsigned char *buff, em_haul_ty
     // connection type flags    
     attr = reinterpret_cast<data_elem_attr_t *> (tmp);
     attr->id = htons(attr_id_conn_type_flags);
-    size = sizeof(unsigned short);
+    size = EM_CONN_TYPE_FLAGS_LEN;
     attr->len = htons(size);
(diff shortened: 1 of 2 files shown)


src/em/config/em_configuration.cpp
```diff
@@ -2148,7 +2150,8 @@ unsigned short em_configuration_t::create_m2_msg(unsigned char *buff, em_haul_ty
     // os version   
     attr = reinterpret_cast<data_elem_attr_t *> (tmp);
     attr->id = htons(attr_id_os_version);
-    size = sizeof(unsigned short);
+    // Os version should be 4 bytes
+    size = 4;
     attr->len = htons(size);
     memcpy(attr->val, &rf_band, sizeof(attr->val));
```
amarnathhullur: The sizeof(attr->val)) would return 0 since the definition of attr->val is unsigned char val[0], so we don't seem to be copying anything in attr->val but just incrementing by 4 bytes.
This seems like wrong code earlier and needs to be changed where ever sizeof(attr->val) is being used for memcpy as it would not copy anything to attr->val

amarnathhullur (changes requested):

amarnathhullur (approved):

PR #528, opened 2025-12-11
syncing main with develop

kpandu987:
syncing main with develop for RDK release


kpandu987:
this is not needed.. will raise new  PR

PR #537, opened 2025-12-17
RDKBACCL-1334:Interop - Encrypted settings TLV in M2-WSC is not padded as per spec

akshaya3495:
[Changes]
Enabled padding in encryption and decryption API's to support PKCS#7 padding method.


akshaya3495:

PR #543, opened 2025-12-19
RDKBACCL-1334: Interop - Encrypted settings TLV in M2-WSC is not padded as per spec (#537)

akshaya3495:
Enabled padding in encryption and decryption APIs to support PKCS#7 padding method.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -36,3 +36,7 @@ Licensed under the FSF Unlimited License (with License Retention)
 
 Copyright 2025 RDK Management
 Licensed under the Apache License, Version 2.0
+
+The Data_Elements_JSON_Schema_v3.0.json file is:
+Copyright (C) 2021 Wi-Fi Alliance. All Rights Reserved.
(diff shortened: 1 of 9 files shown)


inc/em_crypto.h
```diff
@@ -289,11 +289,11 @@ class em_crypto_t {
 	 * @param[in] data_len Length of input ciphertext.
 	 * @param[in] disable_padding If true, disables padding (default is true).
 	 *
-	 * @return 1 on success, 0 on failure.
+	 * @return 1 or length of plain data on success, 0 on failure.
```
amarnathhullur: Instead of returning 1 or length on success, why don't we return just the length on success, positive value indicates success. 
For failure can return 0. 
Ensure that functions calling this and checking for 1 explicitly for success is changed.

amarnathhullur (changes requested):
There seems to be build error with this change. Please check and update on that and also ensure that there are 0 warnings with these changes when compiled in RPI.

PR #545, opened 2025-12-22
RDKBACCL-1334:Interop - Encrypted settings TLV in M2-WSC is not padded as per spec (#545)

akshaya3495:
Enabled padding in encryption and decryption API's to support PKCS#7 padding method.


amarnathhullur (approved):
Looks ok to me

amarnathhullur (approved):

PR #546, opened 2025-12-23
RDKBACCL-1365: Interop - SSID's are not configured with security.

dhiyanesh-dhiyanesh:
Reason for change:
1.Auth type is received wrong due to byte swap added htons and fixed. 
2.Encryption type tlv is now part of WSC M2 message.

### DIFF
--- a/src/em/config/em_configuration.cpp
+++ b/src/em/config/em_configuration.cpp
@@ -4659,20 +4659,21 @@ int em_configuration_t::handle_agent_list_msg(uint8_t *buff, unsigned int len, u
 
 int em_configuration_t::create_encrypted_settings(unsigned char *buff, em_haul_type_t haul_type)
 {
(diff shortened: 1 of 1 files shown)


dhiyanesh-dhiyanesh:

src/em/config/em_configuration.cpp
```diff
@@ -4738,15 +4733,29 @@ int em_configuration_t::create_encrypted_settings(unsigned char *buff, em_haul_t
     tmp += (sizeof(data_elem_attr_t) + size);
 
     // auth type
+    auth_type = EM_AUTH_WPA2PSK | EM_AUTH_SAE_AKM8;
```
amarnathhullur, line 4736: Can the auth_type be changed to have only WPA3 personal i.e. enable only EM_AUTH_SAE_AKM8

PR #555, opened 2026-01-05
RDKBACCL-1193: APMLD and AffiliatedAP for TR-181 added

meycint:
Reason for change: Getters for many objects and parameters are missing for TR-181.
Risks: Low


src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -2585,12 +2618,32 @@ bus_error_t dm_easy_mesh_ctrl_t::device_tget_inner(char *event_name, raw_data_t
...
-        if (sec_cap == NULL) {
-            em_printfout("NULL sec_cap");
-            return bus_error_invalid_input;
+        ++device_cnt;
+        dm = dm_ctrl->get_next_dm(dm);
+    }
+
+    /* Iterate according to dm id */
+    for (unsigned int idx = 1, cnt = 0; cnt < device_cnt; idx++) {
+        dm = dm_ctrl->get_first_dm();
```
amarnathhullur, line 2627: Want to confirm that in this case it is always starting from the first_dm. This I presume is to get the right dm. Is this logic tested with multiple extender? What is the amount of testing addressed with this change?

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -4227,6 +4260,374 @@ bus_error_t dm_easy_mesh_ctrl_t::sta_tget_params(dm_easy_mesh_t *dm, const char
...
+        dm_ctrl->property_append_tail(property, root, idx, "APMLDConfig.EMLMREnabled", ami->emlmr);
+        dm_ctrl->property_append_tail(property, root, idx, "APMLDConfig.EMLSREnabled", ami->emlsr);
+        dm_ctrl->property_append_tail(property, root, idx, "APMLDConfig.STREnabled", ami->str);
+        dm_ctrl->property_append_tail(property, root, idx, "APMLDConfig.NSTREnabled", ami->nstr);
+
+        snprintf(path, sizeof(path) - 1, "%s%d.AffiliatedAP.", root, idx);
+        dm_ctrl->affap_tget_params(dm, path, ami, property);
+
+        //snprintf(path, sizeof(path) - 1, "%s%d.STAMLD.", root, idx);
+        //dm_ctrl->stamld_tget_params(dm, path, property);
```
amarnathhullur, line 4413: Is the above commented code added for future enablement?

amarnathhullur (changes requested):

src/ctrl/dm_easy_mesh_ctrl.cpp
meycint: Yes, and I am currently working on that part.

src/ctrl/dm_easy_mesh_ctrl.cpp
meycint: This statement gets the first dm_easy_mesh_t object of the dm_easy_mesh_ctrl_t object. In other words the first item, by calling hash_map_get_first(), from the m_list (type hash_map_t) object of dm_easy_mesh_list_t. So it must retrieve the first one. The code snippet (with following 6 lines) here tries to find the device having the device instance idx.

amarnathhullur (approved):
Changes looks good

amarnathhullur (approved):

amarnathhullur (approved):

PR #558, opened 2026-01-09
RDKBACCL-1365: Interop - SSID's are not configured with security.

dhiyanesh-dhiyanesh:
Reason for change:
1.Auth type is received wrong due to byte swap added htons and ntohs to fix. 
2.Encryption type tlv is now part of WSC M2 message.


amarnathhullur:
This change has a dependency on change in wifi_easymesh_translator.c which is addressed as part of PR - https://github.com/rdkcentral/OneWifi/pull/824

amarnathhullur (approved):

PR #560, opened 2026-01-12
RDKBACCL-1411 : Misconfiguration of SSID on Agent

AbijithNayak:
triggering renew process when there is a misconfigure of ssid in topology response

### DIFF
--- a/src/em/config/em_configuration.cpp
+++ b/src/em/config/em_configuration.cpp
@@ -37,7 +37,7 @@
 #include <unistd.h>
 #include <pthread.h>
 #include <openssl/rand.h>
-#include <assert.h>
+#include <ctime>
 #include <string>
 #include <vector>
 #include <algorithm>
@@ -58,6 +58,10 @@
 unsigned short em_configuration_t::msg_id = 0;
 // OUI value of Comcast
 static const unsigned char em_vendor_oui[EM_VENDOR_OUI_SIZE] = {0xd8, 0x9c, 0x8e};
+#define MAX_TOPO_RESP_MISCONF_RECEIVED 3
+#define INCREMENT_TOPO_RESP_COUNT 1
+#define TOPO_RETRY_MAX_TIME_SEC 120
+#define TOPO_RETRY_INTERVAL_SEC 10
 
 /* Extract N bytes (ignore endianess) */
 static inline void _EnB(uint8_t **packet_ppointer, void *memory_pointer, uint32_t n)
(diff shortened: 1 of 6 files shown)


src/orch/em_orch_ctrl.cpp
```diff
@@ -55,6 +56,14 @@ void em_orch_ctrl_t::orch_transient(em_cmd_t *pcmd, em_t *em)
 	
 	switch (pcmd->m_type) {
 		case em_cmd_type_em_config:
+            if ((pcmd->get_orch_op() == dm_orch_type_topo_sync) && (em->get_state() == em_state_ctrl_misconfigured)) {
+                    em_printfout("Orchestration transient:%s(%s), em state:%s\n",
+                                           em_cmd_t::get_orch_op_str(pcmd->get_orch_op()), em_cmd_t::get_cmd_type_str(pcmd->m_type), 
+                                           em_t::state_2_str(em->get_state()));
+                    em_printfout("Canceling cmd: %s because of misconfiguration", pcmd->get_cmd_name());
+                    cancel_command(pcmd->get_type());
+                    ctrl->start_complete();
```
amarnathhullur, line 67: What does ctrl->start_complete is supposed to address?

amarnathhullur (dismissed):

src/orch/em_orch_ctrl.cpp
AbijithNayak: Once the command is cancelled, it address the trigger of the reconfiguration process.

src/em/config/em_configuration.cpp
```diff
@@ -1435,26 +1437,40 @@ int em_configuration_t::handle_ap_operational_bss(unsigned char *buff, unsigned
...
-	
                 memcpy(dm_bss->m_bss_info.bssid.mac, bss->bssid, sizeof(mac_address_t));
                 memcpy(dm_bss->m_bss_info.ruid.mac, radio->ruid, sizeof(mac_address_t));
                 dm->set_num_bss(dm->get_num_bss() + 1);
-			}
-            strncpy(dm_bss->m_bss_info.ssid, bss->ssid, bss->ssid_len);
-			dm_bss->m_bss_info.enabled = true;
-			strncpy(dm_bss->m_bss_info.timestamp, time_date, sizeof(em_long_string_t));
+            }
+            char ssid_buf[32];
```
amarnathhullur: Don't we have a macro for the magic number 32, also SSID max len should it not be 33 to accommodate a \0 at the end

amarnathhullur (changes requested):
Minimal change

amarnathhullur:
This change is not complete. It should address the below retry logic
a) Wait for a max of 120 seconds for the topo response to have all the configured SSID from the agent, 120 second value should be configurable, could be a #define value initially. Go to misconfigured state only after 120 seconds. 
b) If there is no topo response from Agent, send topo query every 1 second till topo response is received. Once topo response is received, the timer can be a 10 second timer to periodically check the topo response whether it has all the configured SSID till a max of 120 seconds.

amarnathhullur:
Replaced by PR - https://github.com/rdkcentral/unified-wifi-mesh/pull/565

PR #561, opened 2026-01-14
RDKBACCL-1393 Add support to enable/disable network profile with respect to band and haultype

prashant-singh1:
handle enable/disable with respect to band also excluded disabled ssid from topo response.

### DIFF
--- a/src/em/config/em_configuration.cpp
+++ b/src/em/config/em_configuration.cpp
@@ -3259,6 +3259,7 @@ int em_configuration_t::create_autoconfig_wsc_m2_msg(unsigned char *buff, unsign
 {
     unsigned short  msg_type = em_msg_type_autoconf_wsc;
     int len = 0;
+    bool is_m2_present = false;
     unsigned int i, num_hauls = em_haul_type_max;
     em_cmdu_t *cmdu;
     em_tlv_t *tlv;
@@ -3323,9 +3324,16 @@ int em_configuration_t::create_autoconfig_wsc_m2_msg(unsigned char *buff, unsign
             em_printfout("skipping M2 for haul_type: %d as it's disable, Return value: %d", i, sz);
             continue;
         }
+        // is_m2_present will be true only if atleast one M2 is created.
+        is_m2_present = true;
         tlv->len = htons(sz);
         tmp += (sizeof(em_tlv_t) + sz);
         len += static_cast<int> (sizeof(em_tlv_t) + sz);
+    }
+
+    if(is_m2_present == false) {
+        // M2 not present for crosponding M1, hence returning.
+        return -1;
     }
 
     // default 8022.1q settings tlv 17.2.49
@@ -5058,6 +5066,10 @@ int em_configuration_t::handle_autoconfig_wsc_m1(unsigned char *buff, unsigned i
     }
 
     sz = static_cast<unsigned int> (create_autoconfig_wsc_m2_msg(msg, ntohs(cmdu->id)));
+
+    if (static_cast<int>(sz) < 0) {
+        return -1;
+    }
 
     if (em_msg_t(em_msg_type_autoconf_wsc, em_profile_type_3, msg, sz).validate(errors) == 0) {
         printf("Autoconfig wsc m2 msg failed validation in tnx end\n");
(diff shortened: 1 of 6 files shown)


src/em/config/em_configuration.cpp
```diff
@@ -1961,15 +1965,42 @@ unsigned short em_configuration_t::create_m2_msg(unsigned char *buff, em_haul_ty
...
     if ((net_ssid_info = get_network_ssid_info_by_haul_type(haul_type)) == NULL) {
         em_printfout("Could not find network ssid information for haul type %d", haul_type);
         return 0;
     }
 
-    // skipping the m2 msg if haultype if diabled
-    if(net_ssid_info->enable == false) {
+    // get the current band for m2 messgae create
+    current_band = get_band();
+    isBandEnabled = false;
```
amarnathhullur, line 1979: This would mean by default the bandEnabled is set to False. 
If the band_val in line#1981 returned is blank then the band would be disabled. 
I think this should be other way around (to be backward compatible) that by default the band should be enabled. If the band is explicitly disabled then only the band should be disabled.

amarnathhullur (changes requested):
Also take care of the change that autoconfig wsc M2 message should not be sent if there are no M2 in the message from controller as The spec clearly indicates (one of more WSC TLV containing M2)

Spec extract below
If the message is sent by the Multi-AP Controller:
▪One AP Radio Identifier TLV (see section 17.2.3).
**▪One or more WSC TLV (see [2]) (containing M2) (see [5]).**
▪Zero or one WSC TLV(see [2]) (containing M8) (see [5]).
▪Zero or one Default 802.1Q Settings TLV (see section 17.2.49).
▪Zero or one Traffic Separation Policy TLV (see section 17.2.50).
▪Zero or one Agent AP MLD Configuration TLV (see section 17.2.96)
▪Zero or one Backhaul STA MLD Configuration TLV (see section 17.2.97)

prashant-singh1:
> Also take care of the change that autoconfig wsc M2 message should not be sent if there are no M2 in the message from controller as The spec clearly indicates (one of more WSC TLV containing M2)
> 
> Spec extract below If the message is sent by the Multi-AP Controller: ▪One AP Radio Identifier TLV (see section 17.2.3). **▪One or more WSC TLV (see [2]) (containing M2) (see [5]).** ▪Zero or one WSC TLV(see [2]) (containing M8) (see [5]). ▪Zero or one Default 802.1Q Settings TLV (see section 17.2.49). ▪Zero or one Traffic Separation Policy TLV (see section 17.2.50). ▪Zero or one Agent AP MLD Configuration TLV (see section 17.2.96) ▪Zero or one Backhaul STA MLD Configuration TLV (see section 17.2.97)

As discussed, updated the changes to skip autoconfig message for respective M1, if there is no M2 is present.

src/em/config/em_configuration.cpp
```diff
@@ -3292,11 +3324,18 @@ int em_configuration_t::create_autoconfig_wsc_m2_msg(unsigned char *buff, unsign
...
+        // is_m2_present will be true only if atleast one M2 is created.
+        is_m2_present = true;
         tlv->len = htons(sz);
         tmp += (sizeof(em_tlv_t) + sz);
         len += static_cast<int> (sizeof(em_tlv_t) + sz);
     }
 
+    if(is_m2_present == false) {
+        // M2 not present for crosponding M1, hence returning.
+        return -1;
```
amarnathhullur: you can return 0 here and check for 0 in handle_autconfig.
Also, add a print here indicating that there is no M2 to be added.

src/em/config/em_configuration.cpp
```diff
@@ -3292,11 +3324,18 @@ int em_configuration_t::create_autoconfig_wsc_m2_msg(unsigned char *buff, unsign
...
         }
+        // is_m2_present will be true only if atleast one M2 is created.
+        is_m2_present = true;
         tlv->len = htons(sz);
         tmp += (sizeof(em_tlv_t) + sz);
         len += static_cast<int> (sizeof(em_tlv_t) + sz);
     }
 
+    if(is_m2_present == false) {
+        // M2 not present for crosponding M1, hence returning.
```
amarnathhullur: Typo in the comment change to corresponding

src/em/config/em_configuration.cpp
```diff
@@ -5028,6 +5067,10 @@ int em_configuration_t::handle_autoconfig_wsc_m1(unsigned char *buff, unsigned i
 
     sz = static_cast<unsigned int> (create_autoconfig_wsc_m2_msg(msg, ntohs(cmdu->id)));
 
+    if (static_cast<int>(sz) < 0) {
```
amarnathhullur: Check for 0 here which would not require the static_cast.

amarnathhullur (changes requested):

amarnathhullur (approved):

PR #577, opened 2026-01-30
fixed wrong ieee1905 executable name (#574)

amarnathhullur:
The ieee1905 binary name changed from ieee1905-core to ieee1905. Updating the procd script to be insync with that


amarnathhullur:
Not required now

PR #578, opened 2026-02-04
set handler implementation

aoprasad:


amarnathhullur:
As per the WFA data model there is no write operation of SSID, there is only read operation of SSID, check the link below:
[link](https://usp-data-models.broadband-forum.org/tr-181-2-18-0-usp#D.Device:2.Device.WiFi.DataElements.Network.SSID. )

There is a separate ASYNC Method Data Element [setSSID](https://usp-data-models.broadband-forum.org/tr-181-2-18-0-usp#D.Device:2.Device.WiFi.DataElements.Network.SetSSID) , 
which is used to modify the SSID, passphrase etc, which is under implementation and we should have PR for review soon.

amarnathhullur:
Closing this PR as this change needs to be insync with the latest setSSID change.

PR #585, opened 2026-02-12
RDKBWIFI-337 : In some cases, RDK controller doesnt send topology Query after configuration is successful on all radios

SachinMChougale:
[ISSUE]
    A second M1/M2 exchange on 2.4GHz regresses the per band state machine back to em_state_ctrl_m2_sent,
    so the controller never reaches/maintains the topo_synch_pending. Because of thisTopology Query is never sent.

[Solution]
    Added changes to move state machine to topo_sync_pending if it brought
    back to m2_sent

### DIFF
--- a/src/orch/em_orch_ctrl.cpp
+++ b/src/orch/em_orch_ctrl.cpp
@@ -185,6 +185,14 @@ bool em_orch_ctrl_t::is_em_ready_for_orch_fini(em_cmd_t *pcmd, em_t *em)
                 return true;
             } else if (em->get_state() == em_state_ctrl_configured) {
                 return true;
+            } else if (em->get_state() == em_state_ctrl_wsc_m2_sent) {
+                // A second M1/M2 exchange on 2.4GHz regresses the per band state machine back to em_state_ctrl_m2_sent,
+                // so the controller never reaches/maintains the topo_synch_pending.
+                // Because of thisTopology Query is never sent.
+                // Added changes to move state machine to topo_sync_pending if it brought
+                // back to m2_sent
+                em->set_state(em_state_ctrl_topo_sync_pending);
+                return false;
             }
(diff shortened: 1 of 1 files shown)


src/orch/em_orch_ctrl.cpp
```diff
@@ -185,6 +185,9 @@ bool em_orch_ctrl_t::is_em_ready_for_orch_fini(em_cmd_t *pcmd, em_t *em)
                 return true;
             } else if (em->get_state() == em_state_ctrl_configured) {
                 return true;
+            } else if (em->get_state() == em_state_ctrl_wsc_m2_sent) {
```
amarnathhullur, line 188: 1) Can you add additional comment in this newly added if case: 
why we are doing this change of setting the state to topo_sync_pending and which case is being handled.
2) Will this case be hit only in the  case of the flow where M1 is retransmitted or in any other case?

amarnathhullur (changes requested):

src/orch/em_orch_ctrl.cpp
Nikita-Hakai, line 188: Hi @cm-sachin, 
- if we are getting an M1 again, it should trigger em_bus_event_type_m2_tx event (in handle_autoconfig_wsc_m1()) again and new cmd would be created and em_config would happen again. Is that what is happening and still getting blocked by not going to **topo_sync_pending**. Maybe you can check with below point?
- In **em_configuration_t::process_ctrl_state()**, for the case **em_state_ctrl_topo_sync_pending**, in this scenario the check if all radios are in topo sync state should maybe be changed to 
**_if (em->get_state() >= em_state_ctrl_topo_sync_pending)_**
as M1 might be recvd for one radio and current check would prevent as other radios/ems will already be in state => em_state_ctrl_topo_sync_pending/already configured.

SachinMChougale: Added comment explaining why this change is needed

amarnathhullur (approved):

PR #595, opened 2026-02-19
RDKBWIFI-354 : Send policy config only if all radios are in policy config pending state

SachinMChougale:
Currently, the system sends a policy configuration request as soon as any one radio finishes channel selection. Its better practice to have request sent out after all radios have:
 1. Completed channel selection, and
 2. Moved to the policy_config_pending state.

Changes behaviour:
- The system should wait for all radios to complete channel selection.
- Only when every radio is in the policy_config_pending state should the policy configuration request be sent.

### DIFF
--- a/src/em/policy_cfg/em_policy_cfg.cpp
+++ b/src/em/policy_cfg/em_policy_cfg.cpp
@@ -721,8 +721,61 @@ void em_policy_cfg_t::process_ctrl_state()
 {
     switch (get_state()) {
 		case em_state_ctrl_set_policy_pending:
-        	send_policy_cfg_request_msg();
-			set_state(em_state_ctrl_configured);
+            {
+                em_printfout("Processing set policy pending state for radio %s", util::mac_to_string(get_radio_interface_mac()).c_str());
+                std::vector<em_t *> em_radios;
+                dm_easy_mesh_t *dm = get_data_model();
+
+                get_mgr()->get_all_em_for_al_mac(dm->get_agent_al_interface_mac(), em_radios);
+                for (auto &em : em_radios)
+                {
+                    // Check for null em pointer in vector
+                    if (em == NULL) {
+                        em_printfout("Warning: Null em pointer in vector, skipping");
+                        continue;
+                    }
+                    if (em->get_state() != em_state_ctrl_set_policy_pending)
+                    {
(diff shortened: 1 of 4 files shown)


src/em/policy_cfg/em_policy_cfg.cpp
```diff
@@ -708,8 +707,61 @@ void em_policy_cfg_t::process_ctrl_state()
...
+                for (auto &em : em_radios)
+                {
+                    // Check for null em pointer in vector
+                    if (em == NULL) {
+                        em_printfout("Warning: Null em pointer in vector, skipping");
+                        continue;
+                    }
+                    if (em->get_state() != em_state_ctrl_set_policy_pending)
+                    {
+                        em_printfout("radio %s is in state:%d, not in em_state_ctrl_set_policy_pending, wait for all radios to be in em_state_ctrl_set_policy_pending before sending policy config request",
```
amarnathhullur: for printing the state in string format, use the function em_t::state_2_str(em->get_state())

src/em/policy_cfg/em_policy_cfg.cpp
```diff
@@ -708,8 +707,61 @@ void em_policy_cfg_t::process_ctrl_state()
...
+                        em_printfout("Error: Failed to send policy config request message");
+                        return;
+                    } else {
+                        em_printfout("Policy config request sent successfully, bytes: %d", send_result);
+                    }
+
+                    // Set state for AL node to configured
+                    al_em->set_state(em_state_ctrl_configured);
+                    em_printfout("Set AL node state to configured");
+                    // Set state for all radios to policy config pending
```
amarnathhullur: Comment to be corrected, the state is set to configured state.

amarnathhullur (changes requested):

src/em/policy_cfg/em_policy_cfg.cpp
SachinMChougale: addressed

src/em/policy_cfg/em_policy_cfg.cpp
SachinMChougale: addressed

amarnathhullur (approved):

PR #602, opened 2026-02-26
RDKBWIFI-334: Channel Scan Report not generated properly

Sangeetha-selvaraj18:
Currently:
1. Channel Scan report was generating a FULL Scan report or for wrong Op class and Channel combination.
2. Channel Scan Request is sent for all the radios for each radio.

Changes behaviour:

1. Channel Scan report will be generated only the for the Op class and Channel that controller requested.
2. Controller sends only the specific Opclass and channel set applicable to the respective radio’s frequency band.

### DIFF
--- a/src/em/channel/em_channel.cpp
+++ b/src/em/channel/em_channel.cpp
@@ -175,8 +175,8 @@ short em_channel_t::create_channel_scan_req_tlv(unsigned char *buff)
 
 		req_op_class->op_class = static_cast<unsigned char> (opclass->m_op_class_info.op_class);
 		req_op_class->num_channels = static_cast<unsigned char> (opclass->m_op_class_info.num_channels);
-		unsigned int band = dm_easy_mesh_t::get_freq_band_by_op_class(opclass->m_op_class_info.op_class);
-		unsigned int cur_band = get_band();
+		em_freq_band_t band = dm_easy_mesh_t::get_freq_band_by_op_class(static_cast<int>(opclass->m_op_class_info.op_class));
+		em_freq_band_t cur_band = get_band();
 
 		if (cur_band == band) {
 		    req_op_class->op_class = static_cast<unsigned char> (opclass->m_op_class_info.op_class);
@@ -500,7 +500,7 @@ int em_channel_t::send_channel_scan_report_msg(unsigned int *last_index)
     unsigned int len = 0;
     em_cmdu_t *cmdu;
     em_tlv_t *tlv;
-    unsigned short sz = 0;
+    short sz = 0;
     unsigned char *tmp = buff;
     unsigned short type = htons(ETH_P_1905);
     char date_time[EM_DATE_TIME_BUFF_SZ];
@@ -538,7 +538,7 @@ int em_channel_t::send_channel_scan_report_msg(unsigned int *last_index)
     tlv = reinterpret_cast<em_tlv_t *> (tmp);
     tlv->type = em_tlv_type_timestamp;
 	util::get_date_time_rfc3399(date_time, sizeof(date_time));
-	sz = static_cast<short int> (strlen(date_time) + 1);
+	sz = static_cast<short> (strlen(date_time) + 1);
 	time_len = static_cast<unsigned char> (strlen(date_time));
 	memcpy(tlv->value, &time_len, sizeof(unsigned char));
 	memcpy(tlv->value + 1, date_time, static_cast<size_t> (sz));
--- a/src/em/channel/em_channel.cpp
+++ b/src/em/channel/em_channel.cpp
@@ -175,6 +175,15 @@ short em_channel_t::create_channel_scan_req_tlv(unsigned char *buff)
 
 		req_op_class->op_class = static_cast<unsigned char> (opclass->m_op_class_info.op_class);
 		req_op_class->num_channels = static_cast<unsigned char> (opclass->m_op_class_info.num_channels);
+		unsigned int band = dm_easy_mesh_t::get_freq_band_by_op_class(opclass->m_op_class_info.op_class);
+		unsigned int cur_band = get_band();
+
+		if (cur_band == band) {
+		    req_op_class->op_class = static_cast<unsigned char> (opclass->m_op_class_info.op_class);
+		    req_op_class->num_channels = static_cast<unsigned char> (opclass->m_op_class_info.num_channels);
+		} else {
+		    continue;
+		}
 
 		for (j = 0; j < req_op_class->num_channels; j++) {
 			req_op_class->channel_list[j] = static_cast<unsigned char> (opclass->m_op_class_info.channels[j]);
@@ -184,6 +193,8 @@ short em_channel_t::create_channel_scan_req_tlv(unsigned char *buff)
 		
 		req->num_op_classes++;
 		req_op_class = reinterpret_cast<em_channel_scan_req_op_class_t *> (reinterpret_cast<unsigned char *> (req_op_class) + sizeof(em_channel_scan_req_op_class_t) + req_op_class->num_channels*sizeof(unsigned char));
+		//Do not loop if we get the current radio
+		break;
 	}
 
 	return len;
@@ -313,12 +324,12 @@ int em_channel_t::send_channel_scan_request_msg()
     tmp += (sizeof (em_tlv_t));
     len += (sizeof (em_tlv_t));
     if (em_msg_t(em_msg_type_channel_scan_req, em_profile_type_3, buff, len).validate(errors) == 0) {
-        printf("Channel Selection Request msg failed validation in tnx end\n");
+        em_printfout("Channel Scan Request msg failed validation in tnx end");
         //return -1;
     }
 
     if (send_frame(buff, len)  < 0) {
-        printf("%s:%d: Channel Selection Request msg failed, error:%d\n", __func__, __LINE__, errno);
+        em_printfout("Channel Scan Request msg failed, error:%d", errno);
         return -1;
     }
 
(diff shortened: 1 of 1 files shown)


src/em/channel/em_channel.cpp
```diff
@@ -313,12 +324,12 @@ int em_channel_t::send_channel_scan_request_msg()
     tmp += (sizeof (em_tlv_t));
     len += (sizeof (em_tlv_t));
     if (em_msg_t(em_msg_type_channel_scan_req, em_profile_type_3, buff, len).validate(errors) == 0) {
-        printf("Channel Selection Request msg failed validation in tnx end\n");
+        printf("Channel Scan Request msg failed validation in tnx end\n");
```
amarnathhullur: change printf to em_printfout and remove \n at the end

src/em/channel/em_channel.cpp
```diff
@@ -313,12 +324,12 @@ int em_channel_t::send_channel_scan_request_msg()
...
         //return -1;
     }
 
     if (send_frame(buff, len)  < 0) {
-        printf("%s:%d: Channel Selection Request msg failed, error:%d\n", __func__, __LINE__, errno);
+        printf("%s:%d: Channel Scan Request msg failed, error:%d\n", __func__, __LINE__, errno);
```
amarnathhullur: change printf to em_printfout and remove \n at the end

src/em/channel/em_channel.cpp
```diff
@@ -184,6 +193,8 @@ short em_channel_t::create_channel_scan_req_tlv(unsigned char *buff)
 		
 		req->num_op_classes++;
 		req_op_class = reinterpret_cast<em_channel_scan_req_op_class_t *> (reinterpret_cast<unsigned char *> (req_op_class) + sizeof(em_channel_scan_req_op_class_t) + req_op_class->num_channels*sizeof(unsigned char));
+		//Do not loop if we get the current radio
+		break;
```
amarnathhullur, line 197: What is the reason for this change. Can you please elaborate more?

amarnathhullur (changes requested):
Few comments. Is this change ready for review as the change was still in Drafts. 
Also, what is the testing done with this change?

src/em/channel/em_channel.cpp
Sangeetha-selvaraj18: Hi @amarnathhullur thanks for reviewing.

channel scan request was sent as a set as below:
       Op Class : 81   Channels : 3,6,9  
       Op Class:115  Channels: 36,40,44,48,149,153,157,161,165
       Op Class: 135  Channels:1 
This is sent for all the 3 radios, everytime. This is not necessary. 

As the request is sent everytime the radio is up and configured, I have changed in such a way that, Op class and channel set will be sent only for the particular radio which is raising his request. 
For 2G- Op class:81 Channels: 3,6,9 alone will be sent.

Uploaded the images before and after fix for your reference,

Image

Image

Sangeetha-selvaraj18:
> Few comments. Is this change ready for review as the change was still in Drafts. Also, what is the testing done with this change?

Hi @amarnathhullur 

The fix was tested in Openwrt and easymesh on Banana PI. Since, most of the data in Channel Scan req/report is hardcoded, I have started this is an initial phase of changes to correct the req & response. Tested in colocated agent, and verified that channel scan req and response is sent and received correctly with all the nighbor data.
Validation of each field in the report like BSS load element, will be taken up in a separate ticket.

src/em/channel/em_channel.cpp
```diff
@@ -175,6 +175,15 @@ short em_channel_t::create_channel_scan_req_tlv(unsigned char *buff)
 
 		req_op_class->op_class = static_cast<unsigned char> (opclass->m_op_class_info.op_class);
 		req_op_class->num_channels = static_cast<unsigned char> (opclass->m_op_class_info.num_channels);
+		unsigned int band = dm_easy_mesh_t::get_freq_band_by_op_class(opclass->m_op_class_info.op_class);
+		unsigned int cur_band = get_band();
+
+		if ((band == em_freq_band_60 && cur_band ==  em_freq_band_6) || (cur_band == band)) {
```
amarnathhullur: Based on the latest checkin https://github.com/rdkcentral/unified-wifi-mesh/commit/24547063c625fb32fc170c3debb763801df1e1e3, do you think the first check before the || is required?

src/em/channel/em_channel.cpp
```diff
@@ -313,12 +324,12 @@ int em_channel_t::send_channel_scan_request_msg()
...
     len += (sizeof (em_tlv_t));
     if (em_msg_t(em_msg_type_channel_scan_req, em_profile_type_3, buff, len).validate(errors) == 0) {
-        printf("Channel Selection Request msg failed validation in tnx end\n");
+        em_printfout("Channel Scan Request msg failed validation in tnx end");
         //return -1;
     }
 
     if (send_frame(buff, len)  < 0) {
-        printf("%s:%d: Channel Selection Request msg failed, error:%d\n", __func__, __LINE__, errno);
+        em_printfout("%s:%d: Channel Scan Request msg failed, error:%d", __func__, __LINE__, errno);
```
amarnathhullur: No need of __func__, __LINE__ in case of em_printfout.

src/em/channel/em_channel.cpp
```diff
@@ -471,7 +482,7 @@ short em_channel_t::create_channel_scan_res_tlv(unsigned char *buff, unsigned in
 	}			
 
 	memcpy(tmp, &scan_res->m_scan_result.aggr_scan_duration, sizeof(unsigned int));
-	len += static_cast<short unsigned int> (sizeof(unsigned char));
+	len += static_cast<short unsigned int> (sizeof(unsigned int));
```
amarnathhullur, line 485: good catch

src/em/channel/em_channel.cpp
```diff
@@ -567,7 +578,7 @@ int em_channel_t::send_channel_scan_report_msg(unsigned int *last_index)
     //}
 
     if (send_frame(buff, len)  < 0) {
-        printf("%s:%d: Channel Selection Request msg failed, error:%d\n", __func__, __LINE__, errno);
+        em_printfout("%s:%d: Channel Scan Report  msg failed, error:%d\n", __func__, __LINE__, errno);
```
amarnathhullur: No need of __func__, __LINE__

src/em/channel/em_channel.cpp
```diff
@@ -536,21 +548,20 @@ int em_channel_t::send_channel_scan_report_msg(unsigned int *last_index)
...
-    	tmp += (sizeof(em_tlv_t) + static_cast<short unsigned int> (sz));
-    	len += static_cast<unsigned int> (sizeof(em_tlv_t) + static_cast<short unsigned int> (sz));
-
-		if (len > EM_MAX_CHANNEL_SCAN_RPRT_MSG_LEN) {
-			break;	
-		}
+    for (i = start_idx; i < dm->get_num_scan_results(); i++) {
+        scan_res = dm->get_scan_result(i);
+        if(memcmp(get_radio_interface_mac(), scan_res->m_scan_result.id.scanner_mac, sizeof(mac_address_t)) == 0) { 
+	    tlv = reinterpret_cast<em_tlv_t *> (tmp);
```
amarnathhullur: indentation is incorrect

src/em/channel/em_channel.cpp
```diff
@@ -536,21 +548,20 @@ int em_channel_t::send_channel_scan_report_msg(unsigned int *last_index)
...
+            tlv->type = em_tlv_type_channel_scan_rslt;
+            sz = create_channel_scan_res_tlv(tlv->value, i);
+            tlv->len = htons(static_cast<short unsigned int> (sz));
+
+	    tmp += (sizeof(em_tlv_t) + static_cast<short unsigned int> (sz));
+	    len += static_cast<unsigned int> (sizeof(em_tlv_t) + static_cast<short unsigned int> (sz));
```
amarnathhullur: both of the above line is not indented appropriately

amarnathhullur: Also, you have removed the check 
if (len > EM_MAX_CHANNEL_SCAN_RPRT_MSG_LEN) {
			break;	
		}

What is the reason for that?

amarnathhullur (changes requested):

src/em/channel/em_channel.cpp
Sangeetha-selvaraj18: @amarnathhullur  EM_MAX_CHANNEL_SCAN_RPRT_MSG_LEN is currently defined as 166, which limits the size of a Channel Scan Result TLV in a Channel Scan Report message to a maximum of 166 bytes.

As per my discussion with Rakhil, this restriction was originally introduced because the IEEE layer did not support frame sizes larger than 1024 bytes. Since the IEEE layer now supports fragmentation for messages exceeding this size, we can remove this restriction. Along with this change, I have also increase the buffer size appropriately to ensure that all TLVs can be accommodated.(as done in https://github.com/rdkcentral/unified-wifi-mesh/pull/598)

src/em/channel/em_channel.cpp
Sangeetha-selvaraj18: Corrected

src/em/channel/em_channel.cpp
Sangeetha-selvaraj18: Corrected

src/em/channel/em_channel.cpp
Sangeetha-selvaraj18: Yes, you are right, This commit was merged last friday, after I created a PR. I did a git pull and verified its working correctly. Hence removed that condition here

amarnathhullur (dismissed):

amarnathhullur (approved):

PR #604, opened 2026-02-26
RDKBWIFI-346: vlan id info does not show properly with network topology graph on rdkb cli

prashant-singh1:
Corrected the bsslist encode function to get the vlan from networkssidList
verified the vlan ID showing correctly on network topology graph after the fix.

### DIFF
--- a/src/ctrl/em_network_topo.cpp
+++ b/src/ctrl/em_network_topo.cpp
@@ -66,6 +66,10 @@ void em_network_topo_t::encode(cJSON *parent)
 							m_data_model->m_radio[i].m_radio_info.id.ruid, sizeof(mac_address_t)) == 0) {
 				bss_obj = cJSON_CreateObject();
 				m_data_model->m_bss[j].encode(bss_obj, true);
+				const em_network_ssid_info_t* network_ssid_info = m_data_model->get_network_ssid_info_by_haul_type(m_data_model->m_bss[j].m_bss_info.id.haul_type);
+				if (network_ssid_info != NULL) {
+				    cJSON_AddNumberToObject(bss_obj, "VlanID", network_ssid_info->vlan_id);
+				}
 				sta_list_obj = cJSON_AddArrayToObject(bss_obj, "STAList");
 				dm_easy_mesh_ctrl_t *dm_ctrl = reinterpret_cast<dm_easy_mesh_ctrl_t *>(em_ctrl_t::get_em_ctrl_instance()->get_data_model(GLOBAL_NET_ID));
 				if (dm_ctrl != NULL) {
--- a/src/rdkb-cli/static/script.js
+++ b/src/rdkb-cli/static/script.js
@@ -2215,7 +2215,7 @@ async handleWebSocketMessage(data) {
         .map(bss => {
           const bandLabel = bss.Band === 0 ? '2.4GHz' :
             bss.Band === 1 ? '5GHz' :
-            bss.Band === 2 ? '6GHz' : 'Unknown';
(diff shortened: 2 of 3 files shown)


src/ctrl/em_network_topo.cpp
```diff
@@ -66,6 +66,8 @@ void em_network_topo_t::encode(cJSON *parent)
 							m_data_model->m_radio[i].m_radio_info.id.ruid, sizeof(mac_address_t)) == 0) {
 				bss_obj = cJSON_CreateObject();
 				m_data_model->m_bss[j].encode(bss_obj, true);
+				const em_network_ssid_info_t* network_ssid_info = m_data_model->get_network_ssid_info_by_haul_type(m_data_model->m_bss[j].m_bss_info.id.haul_type);
+				cJSON_AddNumberToObject(bss_obj, "VlanID", network_ssid_info->vlan_id);
```
amarnathhullur: The get_network_ssid_info_by_haul_type can return NULL in case of error scenario. So, check should be done on network_ssid_info before dereferencing.

amarnathhullur (changes requested):

src/ctrl/em_network_topo.cpp
prashant-singh1: Added the NULL check for network_ssid_info

src/rdkb-cli/static/script.js
```diff
@@ -2215,7 +2215,7 @@ async handleWebSocketMessage(data) {
         .map(bss => {
           const bandLabel = bss.Band === 0 ? '2.4GHz' :
             bss.Band === 1 ? '5GHz' :
-            bss.Band === 2 ? '6GHz' : 'Unknown';
+            bss.Band === 3 ? '6GHz' : 'Unknown';
```
amarnathhullur, line 2218: Why 3 is referred as 6Ghz? What is vaue 2?

src/ctrl/em_network_topo.cpp
```diff
@@ -66,6 +66,10 @@ void em_network_topo_t::encode(cJSON *parent)
 							m_data_model->m_radio[i].m_radio_info.id.ruid, sizeof(mac_address_t)) == 0) {
 				bss_obj = cJSON_CreateObject();
 				m_data_model->m_bss[j].encode(bss_obj, true);
+				const em_network_ssid_info_t* network_ssid_info = m_data_model->get_network_ssid_info_by_haul_type(m_data_model->m_bss[j].m_bss_info.id.haul_type);
+				if (network_ssid_info != NULL) {
+				    cJSON_AddNumberToObject(bss_obj, "VlanID", network_ssid_info->vlan_id);
+				}
```
amarnathhullur, line 72: what would be the else case?

amarnathhullur (changes requested):
Some minor question/comments

src/ctrl/em_network_topo.cpp
prashant-singh1: In case of network_ssid_info == NULL, Vlan id will not be added with json response, 
In rdkb cli side, we are already handling if vlan id is not found in response, it will set the vlan id value to 0.
So in else case gui will show vlan id as 0

src/rdkb-cli/static/script.js
prashant-singh1: Previously 
Band value 0 used for 2.4GHz, 1 for 5 Ghz and 2 for 6Ghz, but in some recent changes observed the value of 3 is used  for 6 GHz band, hence in gui, also corrected this.

amarnathhullur (approved):

PR #605, opened 2026-02-26
RDKBWIFI-367 : Race condition of em state while sending policy config request if two em config commands are queued in orchestration for same radio

SachinMChougale:


src/em/policy_cfg/em_policy_cfg.cpp
```diff
@@ -734,7 +734,7 @@ void em_policy_cfg_t::process_ctrl_state()
                         em_printfout("Warning: Null em pointer in vector, skipping");
                         continue;
                     }
-                    if (em->get_state() != em_state_ctrl_set_policy_pending)
+                    if (em->get_state() != em_state_ctrl_set_policy_pending && em->get_state() != em_state_ctrl_configured)
```
amarnathhullur, line 740: This change would make the policy config request to be sent both in set policy pending and configured state. Can you update the comment why this is being done and also the below log indicates not in set_policy_pending, that needs to be changed as well

src/em/policy_cfg/em_policy_cfg.cpp
```diff
@@ -744,7 +744,7 @@ void em_policy_cfg_t::process_ctrl_state()
                 }
                 // If all radios are in set policy pending state, send policy config request on one of them,
                 // ignore sending policy config request on other radios
-                if (!em_radios.empty() && this == em_radios.front())
+                if (!em_radios.empty())
```
amarnathhullur, line 750: With this change will the policy be not sent on all three radios. We should restrict sending only one multi-AP config per device. How is that ensured if this check is removed?

amarnathhullur (changes requested):

src/em/policy_cfg/em_policy_cfg.cpp
SachinMChougale: This change helps in sending policy config request when some radios are already configured state and one or two are in policy_pending state.
This case hits when some radios are reconfigured (exchange M1 and M2 multiple times).
I will update comment!

src/em/policy_cfg/em_policy_cfg.cpp
SachinMChougale: previously we were checking for condition of sending policy config for first radio.
After this change the policy config request will be sent for any one radio em whose orchestration hits and process_control of that radio is triggered.
We are ensuring sending only one policy config per device by setting all radios configured after sending request.

amarnathhullur: With the above if check "if (em->get_state() != em_state_ctrl_set_policy_pending && em->get_state() != em_state_ctrl_configured)" allowing for the em to come to this, will it not send policy config request again?

amarnathhullur (approved):

PR #608, opened 2026-02-27
RDKBACCL-1500 : Need to remove broadcast test lines from rdkb bpi script

kpandu987:
Reason for change: unable to start the ieee1905 systemd service with notify type during bootup
also, added lan0 check to get the mac address for virt_peer interface in absence of erouter0
Test Procedure:Able to launch ieee1905 systemd service with notify type during bootup 
Risks: None

### DIFF
--- a/config/rdkb/banana-pi/setup_veth_for_em.sh
+++ b/config/rdkb/banana-pi/setup_veth_for_em.sh
@@ -44,6 +44,12 @@ if [ "$DO_OBTAIN_MAC" = true ]; then
             echo "Reading mac:$primary_addr from eth0"
             primary_veth0_mac=$(echo $primary_addr | tr -d ':')
             primary_veth0_mac=$((0x$primary_veth0_mac + 0x10))
+    elif [ -e "/sys/class/net/lan0/address" ]; then
+            primary_addr="$(cat /sys/class/net/lan0/address)"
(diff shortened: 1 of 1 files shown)


config/rdkb/banana-pi/setup_veth_for_em.sh
```diff
@@ -39,6 +39,12 @@ if [ "$DO_OBTAIN_MAC" = true ]; then
             #Convert the mac address into hex and increment by 1
             primary_veth0_mac=$(echo $primary_addr | tr -d ':')
             primary_veth0_mac=$((0x$primary_veth0_mac + 0x10))
+    elif [ -e "/sys/class/net/lan0/address" ]; then
```
cpokuru, line 47: For now this is ok.in long run all calls to sdk suggested to be through HAL

kpandu987: ok chandra

amarnathhullur: This will consider lan0 address as the 2nd priority if erouter0 is not present, should this be moved after eth0 giving eth0 2nd priority instead of lan0?

kpandu987: Taken care

amarnathhullur (approved):

PR #617, opened 2026-03-11
RDKBWIFI-380: Updates for Channel Preference TLV processing in Channel selection request on Agent

prathbr:
- Aggregate a list of opclasses and channels from the message TLVs
- Send the list for analysis and activation on the agent DM

Updates in analyze_channel_sel_req to store the message information in DM and activate opclass/channel with best preference
- Invalidate old data if present for the RUID in the DataModel
- Process the aggregated list received from controller and store in the data model
- Select the opclass/channel with best preference in the shared list
- Update/add the best channel to Onewifi Subdoc for further processing and activation

Assumptions:
- A Channel Selection Request contains data for one RUID only
- In case no opclass/channel is present in the message, it shall invalidate existing entries. No update to current channel/onewifi subdoc

### DIFF
--- a/src/agent/dm_easy_mesh_agent.cpp
+++ b/src/agent/dm_easy_mesh_agent.cpp
@@ -396,49 +396,123 @@ int dm_easy_mesh_agent_t::analyze_channel_pref_query(em_bus_event_t *evt, em_cmd
 
 int dm_easy_mesh_agent_t::analyze_channel_sel_req(em_bus_event_t *evt, wifi_bus_desc_t *desc,bus_handle_t *bus_hdl)
 {
-	unsigned int i = 0, noofopclass = 0;
-	op_class_channel_sel *channel_sel;
-	em_op_class_info_t *dm_op_class;
-	em_tx_power_limit_t	*tx_power_limit;
-	em_spatial_reuse_req_t *spatial_reuse_req;
-	em_eht_operations_t *eht_ops;
-	bool found_mesh_sta = false;
-	em_bss_info_t *bss_info;
+    unsigned int i = 0, j = 0, noofopclass = 0;
+    op_class_channel_sel *channel_sel;
+    em_op_class_info_t *dm_op_class;
+    em_tx_power_limit_t *tx_power_limit;
+    em_spatial_reuse_req_t *spatial_reuse_req;
+    em_eht_operations_t *eht_ops;
+    bool found_mesh_sta = false;
+    em_bss_info_t *bss_info;
 
 
-	channel_sel = reinterpret_cast<op_class_channel_sel*> (evt->u.raw_buff);
-	printf("%s:%d No of opclass=%d tx=%d\n", __func__, __LINE__,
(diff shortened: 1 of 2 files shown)


prathbr:

src/agent/dm_easy_mesh_agent.cpp
```diff
@@ -396,49 +396,117 @@ int dm_easy_mesh_agent_t::analyze_channel_pref_query(em_bus_event_t *evt, em_cmd
...
+        // Assumption: recevied channel_sel contains entries for one RUID only
+        if ((dm_op_class->id.type == em_op_class_type_anticipated) &&
+            memcmp(&dm_op_class->id.ruid, &channel_sel->op_class_info[0].id.ruid, sizeof(mac_address_t)) == 0) {
+            dm_op_class->pref_valid = false;
+        }
+    }
+
+    if (channel_sel->num == 0) {
+        // UNEXPECTED: Channel Selection Request will have atleast one OPCLASS in Channel Preference TLV
+        return -1;
```
amarnathhullur, line 431: Add error print in this case before returning -1

amarnathhullur (changes requested):
Need high level overview of this change, will setup a quick sync up for review.

src/agent/dm_easy_mesh_agent.cpp
```diff
@@ -396,49 +396,117 @@ int dm_easy_mesh_agent_t::analyze_channel_pref_query(em_bus_event_t *evt, em_cmd
...
+            dm_op_class = &this->m_op_class[noofopclass].m_op_class_info;
+            memcpy(dm_op_class, &channel_sel->op_class_info[i], sizeof(em_op_class_info_t));
+            dm_op_class->id.type = em_op_class_type_anticipated;
+            dm_op_class->pref_valid = true;
+            noofopclass++;
+        }
+
+        // Check for all channels in the entry to maintian most preferred channel and opclass
+        for (unsigned int ch_idx = 0; ch_idx < dm_op_class->num_channels; ch_idx++) {
+            if (dm_op_class->channel_pref[ch_idx] > highest_anticipated_preference) {
```
amarnathhullur: should there be a check for the channel_pref[ch_idx] to be within certain range before checking for > highest_anticipated_preference?

amarnathhullur (changes requested):

src/agent/dm_easy_mesh_agent.cpp
prathbr: Added error print.

src/agent/dm_easy_mesh_agent.cpp
prathbr: Added a check to ensure ch_idx is within the range of EM_MAX_CHANNELS_IN_LIST. Addressed the type mismatch between highest_anticipated_preference and dm_op_class->chan_pref by changing the declaration of highest_anticipated_preference to unsigned char. Also masked the reported preference with F0 before checking the preference value

amarnathhullur (approved):

PR #618, opened 2026-03-16
Added support for Capabilities.HECapabilities in tr_181 layer.

sriraaman-c:
Reason for change : Have added support on tr_181 side to expose the HE Related capabilities.

### DIFF
--- a/src/ctrl/dm_easy_mesh_ctrl.cpp
+++ b/src/ctrl/dm_easy_mesh_ctrl.cpp
@@ -3368,6 +3368,65 @@ char* dm_easy_mesh_ctrl_t::get_vht_caps_str(em_ap_vht_cap_t *vht, char *buf, siz
     return buf;
 }
 
+char* dm_easy_mesh_ctrl_t::get_he_caps_str(em_ap_he_cap_t *he, char *buf, size_t buf_len)
+{
+    const size_t capped_mcs_len = static_cast<size_t>(he->sprt_mcs_len);
+
+    const size_t payload_len = 1                       // mcs_len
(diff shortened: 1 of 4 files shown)


src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -3362,6 +3362,65 @@ char* dm_easy_mesh_ctrl_t::get_vht_caps_str(em_ap_vht_cap_t *vht, char *buf, siz
...
+    flags1 |= static_cast<uint8_t>((he->su_beamformer_cap  & 0x1) << 7);
+    flags1 |= static_cast<uint8_t>((he->mu_beamformer_cap  & 0x1) << 6);
+    flags1 |= static_cast<uint8_t>((he->ul_mimo_cap        & 0x1) << 5);
+    flags1 |= static_cast<uint8_t>((he->ul_mimo_ofdma_cap  & 0x1) << 4);
+    flags1 |= static_cast<uint8_t>((he->dl_mimo_ofdma_cap  & 0x1) << 3);
+    flags1 |= static_cast<uint8_t>((he->ul_ofdma_cap       & 0x1) << 2);
+    flags1 |= static_cast<uint8_t>((he->dl_ofdma_cap       & 0x1) << 1);
+    flags1 |= static_cast<uint8_t>(he->reserved & 0x1);
+    data.push_back(flags1);
+
```
amarnathhullur, line 3415: how is the values present in data, copied to buf in the #else case?
I don;t see that happening for other cases as well of ht_str and vht_caps.

How have these been verified?

amarnathhullur (changes requested):
Data does not seem to have been copied to buf

PR #621, opened 2026-03-18
RDKBWIFI-378: Redesign of rdkb cli channel change to support channel update per radio and set preference.

prashant-singh1:
The changes include.
  a) Rdkb cli support change channel for global and specific radio.
       for example, 
       if user select 2.4 and device id from drop down the channel change will be applicable for 2.4Ghz radio id of selected device. 
       if user select "ALL Station" the channel config will be global and applicable for all radio ID
b) Option to select the preference for each selected channel.
     - if channel is selected and preference not set, the default preference will be set as 15 else the selected preference from 0 to 14 will be set.

This Changes does not include handling ruid based channel change handling and db update from ctrl side, this need to be handled separately.

### DIFF
--- a/src/dm/dm_easy_mesh.cpp
+++ b/src/dm/dm_easy_mesh.cpp
@@ -1122,7 +1122,7 @@ int dm_easy_mesh_t::decode_config_set_policy(em_subdoc_info_t *subdoc, const cha
 int dm_easy_mesh_t::decode_config_set_channel(em_subdoc_info_t *subdoc, const char *key, unsigned int index, unsigned int *num)
 {
     cJSON *parent_obj, *net_obj, *net_obj_id; 
-	cJSON *target_arr_obj, *target_obj, *channel_arr_obj;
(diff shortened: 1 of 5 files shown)


src/dm/dm_easy_mesh.cpp
```diff
@@ -1194,13 +1194,19 @@ int dm_easy_mesh_t::decode_config_set_channel(em_subdoc_info_t *subdoc, const ch
         	return -1;
 		}
 
+		if ((channel_pref_arry_obj = cJSON_GetObjectItem(target_obj, "ChannelPrefList")) == NULL) {
+			cJSON_Delete(parent_obj);
+			printf("%s:%d: %s not present\n", __func__, __LINE__, target_key);
```
amarnathhullur: printf cannot be used, change to em_printfout.

amarnathhullur (changes requested):

src/dm/dm_easy_mesh.cpp
prashant-singh1: changed to em_printfout

amarnathhullur (approved):

PR #624, opened 2026-03-24
RDKBWIFI-322: Radio cap null fix during dml translation

Nikita-Hakai:


src/dm/dm_easy_mesh.cpp
amarnathhullur: Can you also add check that index not to be >= EM_MAX_BANDS?

amarnathhullur (changes requested):
Minor comment

amarnathhullur (approved):

PR #630, opened 2026-03-30
RDKBWIFI-391 : RDKBCLI - Channel Selection - Controller orchestration…

SachinMChougale:
… state machine not updated/cleaned correctly during channel change

Description:
    History: The issue of handling OCR (Operating Channel Report) during configuration was fixed in #593. While testing onboarding of third-party agents,
             it was observed that OCR was being received before the dm_orch_type_channel_cnf orchestration state became active.
             Due to this, the orchestration command was updated to handle channel selection in a single orchestration to ensure proper state management and cleanup during channel changes.
    
    This fix was not applied for channel selection request when triggered from GUI/any other way apart from em_config.

### DIFF
--- a/src/cmd/em_cmd_set_channel.cpp
+++ b/src/cmd/em_cmd_set_channel.cpp
@@ -45,11 +45,9 @@ em_cmd_set_channel_t::em_cmd_set_channel_t(em_cmd_params_t param, dm_easy_mesh_t
     memset(reinterpret_cast<unsigned char *> (&m_orch_desc[0]), 0, EM_MAX_CMD*sizeof(em_orch_desc_t));
 
     m_orch_op_idx = 0;
-    m_num_orch_desc = 2;
+    m_num_orch_desc = 1;
     m_orch_desc[0].op = dm_orch_type_channel_sel;
     m_orch_desc[0].submit = true;
-    m_orch_desc[1].op = dm_orch_type_channel_cnf;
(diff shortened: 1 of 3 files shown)


src/cmd/em_cmd_set_channel.cpp
```diff
@@ -45,11 +45,9 @@ em_cmd_set_channel_t::em_cmd_set_channel_t(em_cmd_params_t param, dm_easy_mesh_t
     memset(reinterpret_cast<unsigned char *> (&m_orch_desc[0]), 0, EM_MAX_CMD*sizeof(em_orch_desc_t));
 
     m_orch_op_idx = 0;
-    m_num_orch_desc = 2;
+    m_num_orch_desc = 1;
     m_orch_desc[0].op = dm_orch_type_channel_sel;
     m_orch_desc[0].submit = true;
-    m_orch_desc[1].op = dm_orch_type_channel_cnf;
```
amarnathhullur, line 51: why is this op not required? Can you provide more detailed documentation in this regard in the commit text

amarnathhullur (changes requested):

src/cmd/em_cmd_set_channel.cpp
SachinMChougale: I have updated commit message with detailed information of why dm_orch_type_channel_cnf op is not needed.

amarnathhullur (approved):

PR #646, opened 2026-04-07
RDKBWIFI-357: For MultiAP Policy Config Request and Channel Scan Request RDK agent is not sending 1905 ACK.

dhiyanesh-dhiyanesh:
- Send ACK message for both MultiAP Policy Config Request and Channel Scan Request
- Added message ID validation to ensure ACKs are sent for the correct request messages
- Enhanced orchestration logic to handle repeated EM config commands properly
- Updated state machine to wait and transition appropriately after ACK processing

### DIFF
--- a/inc/em_orch.h
+++ b/inc/em_orch.h
@@ -220,6 +220,16 @@ public:
 	void cancel_command(em_cmd_type_t type);
 
 	/**!
+	 * @brief Removes em_config command for the EM with radio mac.
+	 *
+	 * This function is used to remove cmd_type_em_config currently active and queued for the EM with the radio mac.
+	 *
+	 * @param[in] radio_mac Radio MAC of the EM for which the command should be removed.
+	 *
+	 */
+	void remove_em_config_cmd_for_em(unsigned char *radio_mac);
+
+	/**!
 	 * @brief Resets the command time for a specific command type.
 	 *
 	 * This function is responsible for resetting the timer associated with a specific command type in the command map.
--- a/src/em/channel/em_channel.cpp
+++ b/src/em/channel/em_channel.cpp
@@ -95,9 +95,9 @@ short em_channel_t::create_channel_pref_tlv_agent(unsigned char *buff, unsigned 
     return len;
 }
 
-int em_channel_t::send_1905_ack_message(unsigned short msg_id)
+int em_channel_t::send_1905_ack_message(unsigned short msg_id, bool ack_from)
 {
-    unsigned char buff[MAX_EM_BUFF_SZ];
+    unsigned char buff[MAX_EM_BUFF_SZ] = {0};
     char *errors[EM_MAX_TLV_MEMBERS] = {0};
     unsigned short  msg_type = em_msg_type_1905_ack;
     size_t len = 0;
@@ -107,17 +107,31 @@ int em_channel_t::send_1905_ack_message(unsigned short msg_id)
     unsigned short type = htons(ETH_P_1905);
     dm_easy_mesh_t *dm = get_data_model();
 
-    memcpy(tmp, dm->get_agent_al_interface_mac(), sizeof(mac_address_t));
-    tmp += sizeof(mac_address_t);
-    len += sizeof(mac_address_t);
+    if (ack_from == ACK_FROM_CTRL) {
+	    memcpy(tmp, dm->get_agent_al_interface_mac(), sizeof(mac_address_t));
+	    tmp += sizeof(mac_address_t);
+	    len += sizeof(mac_address_t);
+            
+	    memcpy(tmp, dm->get_ctrl_al_interface_mac(), sizeof(mac_address_t));
+	    tmp += sizeof(mac_address_t);
+	    len += sizeof(mac_address_t);
 
-    memcpy(tmp, dm->get_ctrl_al_interface_mac(), sizeof(mac_address_t));
-    tmp += sizeof(mac_address_t);
-    len += sizeof(mac_address_t);
+	    memcpy(tmp, reinterpret_cast<unsigned char *> (&type), sizeof(unsigned short));
+	    tmp += sizeof(unsigned short);
+	    len += sizeof(unsigned short);
+    } else {
+	    memcpy(tmp, dm->get_ctrl_al_interface_mac(), sizeof(mac_address_t));
+	    tmp += sizeof(mac_address_t);
+	    len += static_cast<unsigned int> (sizeof(mac_address_t));
+            
+	    memcpy(tmp, dm->get_agent_al_interface_mac(), sizeof(mac_address_t));
+	    tmp += sizeof(mac_address_t);
+	    len += static_cast<unsigned int> (sizeof(mac_address_t));
 
-    memcpy(tmp, reinterpret_cast<unsigned char *> (&type), sizeof(unsigned short));
-    tmp += sizeof(unsigned short);
-    len += sizeof(unsigned short);
+	    memcpy(tmp, reinterpret_cast<unsigned char *> (&type), sizeof(unsigned short));
+	    tmp += sizeof(unsigned short);
+	    len += static_cast<unsigned int> (sizeof(unsigned short));
+    }
 
     cmdu = reinterpret_cast<em_cmdu_t *> (tmp);
     memset(tmp, 0, sizeof(em_cmdu_t));
@@ -146,6 +160,7 @@ int em_channel_t::send_1905_ack_message(unsigned short msg_id)
         return 0;
     }
     em_printfout("1905 ACK send success\n");
+
     return static_cast<int> (len);
 }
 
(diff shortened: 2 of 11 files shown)


inc/em_orch.h
```diff
@@ -219,6 +219,16 @@ class em_orch_t {
...
 
+	/**!
+	 * @brief Removes em_config command for the EM with radio mac.
+	 *
+	 * This function is used to remove cmd_type_em_config currently active and queued for the EM with the radio mac.
+	 *
+	 * @param[in] radio_mac Radio MAC of the EM for which the command should be removed.
+	 *
+	 */
+        void remove_em_config_cmd_for_em(unsigned char *radio_mac);
```
amarnathhullur: update indentation

src/em/channel/em_channel.cpp
```diff
@@ -1970,7 +1986,32 @@ int em_channel_t::handle_operating_channel_rprt(unsigned char *buff, unsigned in
...
+        //check for null em pointer in vector
+        if (em == NULL) {
+            em_printfout("Warning: Null em pointer in vector, skipping");
+            continue;
+        }
+
+        if((em->get_state() == em_state_ctrl_channel_scan_pending) && (response_msg_id == em->m_chan_req_msg_id)) {
+            em->set_state(em_state_ctrl_configured);
+            em->m_chan_req_msg_id = 0;
+        }
```
amarnathhullur, line 2013: Would it be prudent to add an info print if the ack is not handled when the state is other than scan_pending or will it be too much of print?

src/em/config/em_configuration.cpp
```diff
@@ -5074,6 +5074,10 @@ int em_configuration_t::handle_autoconfig_wsc_m1(unsigned char *buff, unsigned i
         printf("%s:%d: autoconfig wsc m2 send failed, error:%d\n", __func__, __LINE__, errno);
         return -1;
     }
+
+    em_printfout("%s:%d: Removing previous em_config command for radio " MACSTRFMT "\n", __func__, __LINE__, MAC2STR(get_radio_interface_mac()));
+    static_cast<em_ctrl_t*>(get_mgr())->get_orch()->remove_em_config_cmd_for_em(get_radio_interface_mac());
```
amarnathhullur: Why is this change required? Can you elaborate more on this?

src/em/policy_cfg/em_policy_cfg.cpp
```diff
@@ -693,10 +696,106 @@ int em_policy_cfg_t::handle_policy_cfg_req(unsigned char *buff, unsigned int len
     }
 
     get_mgr()->io_process(em_bus_event_type_set_policy, reinterpret_cast<unsigned char *> (&policy), sizeof(policy));
+    send_1905_ack_message(ntohs(cmdu->id));
 
     return 0;
 }
 
+int em_policy_cfg_t::send_1905_ack_message(unsigned short msg_id)
```
amarnathhullur, line 704: Why does this send_1905_ack_message not have the 2nd parameter? Is the assumption that it is always from AGENT? Maybe the 2nd parameter can be made as default FROM_AGENT and the caller need not change.

amarnathhullur: Also the implementation of this is exactly same as that of em_channel_t::send_1905_ack_message, can both of those be consolidated into one function at the em level?

src/em/policy_cfg/em_policy_cfg.cpp
```diff
@@ -747,35 +851,27 @@ void em_policy_cfg_t::process_ctrl_state()
...
-                    for (auto &em : em_radios)
-                    {
-                        if (em != NULL)
-                        {
-                            em->set_state(em_state_ctrl_configured);
-                            em_printfout("Set radio %s state to configured",
-                                        util::mac_to_string(em->get_radio_interface_mac()).c_str());
-                        }
-                    }
+			 // Send policy config request and check for errors
```
amarnathhullur: indentation error

src/em/policy_cfg/em_policy_cfg.cpp
```diff
@@ -747,35 +851,27 @@ void em_policy_cfg_t::process_ctrl_state()
...
+                             return;
+                         } else {
+                             em_printfout("Policy config request sent successfully, bytes: %d", send_result);
+                         }
+		    }
+		} else if (!em_radios.empty() && em_radios.front() != this) {
+		    em_printfout("Policy config request message already sent by radio %s, not sending again from radio %s",
+                              util::mac_to_string(em_radios.front()->get_radio_interface_mac()).c_str(),
+                              util::mac_to_string(get_radio_interface_mac()).c_str());
+                    set_state(em_state_ctrl_configured);
```
amarnathhullur: Why are we setting the state to configured blindly? The channel selection is still not complete, so setting the state to configured not sure what is the side effect it will have.

src/orch/em_orch.cpp
```diff
@@ -207,10 +207,58 @@ void em_orch_t::cancel_command(em_cmd_type_t type, std::vector<em_t*> &em_radios
         }
     }
 
-    // reset the elapsed time for the command type to avoid triggering timeouts for cancelled commands.
     reset_cmd_time(m_cmd_map, type);
 }
 
+void em_orch_t::remove_em_config_cmd_for_em(unsigned char *radio_mac)
```
amarnathhullur, line 213: What kind of testing is addressed for this function?

src/em/config/em_configuration.cpp
amarnathhullur: Also this line seems to be causing linking error as this file is common across controller and agent and usage of em_ctrl_t  is not resolved in agent built

inc/em_orch.h
dhiyanesh-dhiyanesh: updated

src/em/policy_cfg/em_policy_cfg.cpp
dhiyanesh-dhiyanesh: resolved

src/em/config/em_configuration.cpp
dhiyanesh-dhiyanesh: Resolved linking error by defining virtual function in em_mgr.h which returns null if its agent context as its defined only for controller.

src/em/policy_cfg/em_policy_cfg.cpp
nikunjshah-cap: There are existing definitions of send_1905_ack_message already in use across em_channel.cpp, em_configuration.cpp, and em_steering.cpp, taking slightly different arguments based on their respective use cases. To stay consistent with this existing pattern, we made the following choices:

- Introduced a ACK send function in em_policy_cfg. Since this flow currently only requires ACKs in a single direction, the second parameter was intentionally not included. If bidirectional ACK support is needed in the future, this can be easily extended.
- Updated em_channel::send_1905_ack_message to explicitly support ACKs in both directions by adding a new argument, rather than changing existing call semantics.

While the core logic is largely similar across these implementations, there are minor variations driven by differences in parameters and usage context.

src/em/config/em_configuration.cpp
nikunjshah-cap: This change was driven by an issue observed during interoperability testing.

In a specific scenario, the Agent sends an M1 while the EM is already in the m2_sent state as part of an ongoing em_config orchestration. When the Controller receives this M1, it queues a new em_config command. As a result, once the first em_config completes, the Controller proceeds to execute the newly queued one, even though it is functionally redundant.

This leads to duplicate orchestration for the same radio, which is unnecessary and can cause unintended behavior.

To address this, when an M1 is received again, we now clean up any existing em_config command for that radio and then queue the new em_config command. This ensures that only one active orchestration exists per radio at any given time, avoids duplication, and allows the flow to proceed as expected.

src/orch/em_orch.cpp
nikunjshah-cap: This change was driven by an issue observed during interoperability testing, where em_config orchestration for a radio was getting duplicated.

To validate the fix:

- We reproduced the issue using a third‑party Agent that exhibited this behavior. After the change, the existing em_config orchestration is correctly cleaned up when an M1 is received again, and a new orchestration is queued and executed successfully without duplication.
- We verified RDK‑to‑RDK onboarding multiple times to ensure there was no regression in the standard flow.

These tests confirmed that the orchestration cleanup and re‑queueing logic behaves as expected and does not negatively impact existing flows.

src/em/channel/em_channel.cpp
dhiyanesh-dhiyanesh: addressed

src/em/policy_cfg/em_policy_cfg.cpp
dhiyanesh-dhiyanesh: addressed

amarnathhullur (approved):

PR #647, opened 2026-04-07
RDKCOM-5596: RDKBDEV-3424 Issue 644: Crash in get_next_tlv

Yashu37:
Problem:

    1.Buffer Underflow: The get_next_tlv function crashes when processing truncated or malformed packets because it lacks a boundary check to verify
      if the remaining buff_len can accommodate the 3-octet TLV header (1-octet Type + 2-octet Length).
    2.Unsafe Memory Access: The code performs ntohs(tlv->len) before validating the buffer size. If buff_len < 3, this results in an out-of-bounds read.
    3.Stability Issue: Malformed network traffic can trigger a Denial of Service (DoS) by crashing the component.

    Steps to reproduce:

    1.Initialize a Truncated Buffer: Create a raw memory buffer with a size smaller than sizeof(em_tlv_t).
      Example: Define a buffer of only 1 or 2 octets (representing a packet that was cut off mid-header).
    2.Invoke the Function: Pass this small buffer and its actual length (e.g., buff_len = 2) into get_next_tlv().
    3.Pointer Dereference (The Failure Point): In the original code, the execution hits the line:
      uint16_t current_tlv_size = sizeof(em_tlv_t) + ntohs(tlv->len);
    4.Observe the Crash: Because the system attempts to read 2 bytes for the len field from a buffer that may only have 0 or 1 byte remaining at that offset, it triggers an Out-of-Bounds Read.

    Fixed:

    I added a check to verify that the buffer size is sufficient to hold the TLV header, logging a 'Truncated packet' error and returning NULL if the length is invalid.

### DIFF
--- a/src/em/em_msg.cpp
+++ b/src/em/em_msg.cpp
@@ -324,6 +324,11 @@ em_tlv_t *em_msg_t::get_next_tlv(em_tlv_t* tlv, em_tlv_t* tlvs_buff, unsigned in
     size_t offset = static_cast<size_t>(signed_offset);
     EM_ASSERT_MSG_TRUE(offset < buff_len, NULL, "TLV offset exceeds buffer length");
 
+    if (buff_len < sizeof(em_tlv_t)) {
+        em_printfout("Truncated packet: not enough space for TLV length field");
+        return NULL;
+    }
+
     // Calculate the size of the current TLV (header + data)
     uint16_t current_tlv_size = sizeof(em_tlv_t) + ntohs(tlv->len);
     


src/em/em_msg.cpp
```diff
@@ -324,6 +319,11 @@ em_tlv_t *em_msg_t::get_next_tlv(em_tlv_t* tlv, em_tlv_t* tlvs_buff, unsigned in
     size_t offset = static_cast<size_t>(signed_offset);
     EM_ASSERT_MSG_TRUE(offset < buff_len, NULL, "TLV offset exceeds buffer length");
 
+    if (buff_len < sizeof(em_tlv_t)) {
+            em_printfout("Truncated packet: not enough space for TLV length field\n");
+            return NULL;
+    }
```
amarnathhullur, line 330: Indentation seems incorrect, seems to have more than 4 spaces.
Also remove \n as em_printfout already adds \n

amarnathhullur (changes requested):

src/em/em_msg.cpp
```diff
@@ -319,6 +319,11 @@ em_tlv_t *em_msg_t::get_next_tlv(em_tlv_t* tlv, em_tlv_t* tlvs_buff, unsigned in
     size_t offset = static_cast<size_t>(signed_offset);
     EM_ASSERT_MSG_TRUE(offset < buff_len, NULL, "TLV offset exceeds buffer length");
 
+    if (buff_len < sizeof(em_tlv_t)) {
+            em_printfout("Truncated packet: not enough space for TLV length field\n");
```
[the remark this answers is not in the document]
Yashu37: The check if (buff_len < sizeof(em_tlv_t)) is necessary to safely handle cases where the packet contains only a single TLV with insufficient length (e.g., 1 or 2 bytes) after the Ethernet and 1905 headers. Without this validation, accessing the TLV length field can lead to out-of-bounds memory access.

Additionally, the condition suggested by Copilot (validating the remaining buffer using remaining_len < sizeof(em_tlv_t)) is also required in get_next_tlv(). This specifically addresses scenarios where the packet contains multiple TLVs, such as a valid first TLV (length ≥ 3 bytes) followed by a truncated second TLV (1 or 2 bytes). In such cases, the initial check is not sufficient, and validating the remaining buffer ensures safe parsing of subsequent TLVs.

src/em/em_msg.cpp
```diff
@@ -319,6 +319,11 @@ em_tlv_t *em_msg_t::get_next_tlv(em_tlv_t* tlv, em_tlv_t* tlvs_buff, unsigned in
...
+            return NULL;
+    }
+
     // Calculate the size of the current TLV (header + data)
     uint16_t current_tlv_size = sizeof(em_tlv_t) + ntohs(tlv->len);
```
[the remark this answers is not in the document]
Yashu37: This concern is already handled, as get_next_tlv() returns NULL when remaining_len < sizeof(em_tlv_t), ensuring no insufficient buffer is passed, and get_first_tlv() also includes a header-size check, so truncated buffers cannot cause out-of-bounds access.

src/em/em_msg.cpp
Yashu37: Addressed

Yashu37 (commented):
Addressed

Yashu37:
@amarnathhullur requested you kindly review this changes. Once this merged, it will help in further stability and unit test.

src/em/em_msg.cpp
```diff
@@ -333,6 +333,13 @@ em_tlv_t *em_msg_t::get_next_tlv(em_tlv_t* tlv, em_tlv_t* tlvs_buff, unsigned in
         return NULL; // No more TLVs
     }
 
+    size_t remaining_len = buff_len - offset;
+
+    if (remaining_len < sizeof(em_tlv_t)) {
```
amarnathhullur: I dont think this check is required here as that should be addressed as part of the call of get_first_tlv below at the end of this function

amarnathhullur (changes requested):

src/em/em_msg.cpp
Yashu37: I added the above two boundary checks to handle cases where get_next_tlv() is invoked directly without first calling get_first_tlv(). In such scenarios, the function may operate on invalid or insufficient buffer data, leading to a crash.

I verified this behavior using the following test:

int test_get_next_tlv(void)
{
    em_tlv_t* ret = NULL;

    printf("packet_len = %d\n", packet_len);
    printf("PCAP file 'output.pcap' created successfully\n");

    em_msg_t obj;

    // Step 1: directly point to first TLV (skip 22 bytes)
    em_tlv_t *current_tlv = reinterpret_cast<em_tlv_t*>(packet + 22);

    // Step 2: get next TLV
    ret = obj.get_next_tlv(
            current_tlv,
            reinterpret_cast<em_tlv_t*>(packet + 22),
            packet_len - 22
    );

    return (ret != NULL) ? 0 : -1;
}

During testing, get_next_tlv() crashed when invoked directly on malformed or insufficient TLV data.
These additional checks ensure safe handling of such cases and prevent invalid memory access.

Yashu37: @amarnathhullur I agree with your point. Since get_next_tlv() is always used together with get_first_tlv(), this check is redundant. I have removed it.

Yashu37: @amarnathhullur  All reviews are addressed.

src/em/em_msg.cpp
```diff
@@ -328,6 +333,13 @@ em_tlv_t *em_msg_t::get_next_tlv(em_tlv_t* tlv, em_tlv_t* tlvs_buff, unsigned in
...
+        em_printfout("Truncated packet: not enough space for TLV header");
+        return NULL;
+    }
```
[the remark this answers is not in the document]
Yashu37: Addressed

Nikita-Hakai (approved):

Yashu37:
Hi @amarnathhullur ,  requested you kindly review this changes.

Nikita-Hakai (approved):

Yashu37:
This is the Jira for this changes: RDKBDEV-3424
Link: https://jira.rdkcentral.com/jira/browse/RDKBDEV-3424

Yashu37:
HI @amarnathhullur requested you kindly review this changes. Once this merged, it will help in further stability and unit test.

pradeeptakdas:
Internal ticket for tracking: https://ccp.sys.comcast.net/browse/RDKCOM-2631

Nikita-Hakai (approved):

PR #653, opened 2026-04-13
RDKBWIFI-410: Rework channel selection flow to delegate decision to platform ACS

dkyncu:
Channel selection is delegated to the platform ACS. The EasyMesh Controller
and Agent no longer pick a channel; they only constrain the set of channels
the platform ACS is allowed to use.

CLI/UI -> Controller(all channels + prefs) -> Agent(exclude list) -> OneWifi(route subdoc) -> Platform ACS(pick channel)

This aligns with FTRS EasyMesh+ CORE_AIR_0013 and CORE_AIR_0014: width and opclass
decisions belong to the platform; the Agent's exclude list only reflects the
channels the ACS daemon may not use.

Controller:
The Channel Preference TLV carries every channel in the radio's band with its
preference byte: operable channels keep a non-zero preference, non-operable
channels are encoded with pref_bits = EM_CH_PREF_NON_OPERABLE. This preserves
the agent capability/preference and the CLI anticipated preferences end to end.
create_channel_pref_tlv and create_transmit_power_limit_tlv take an optional
target EM so a single Channel Selection Request can bundle one TLV pair per radio.

Agent:
analyze_channel_sel_req no longer runs any local channel-picking logic.
It now builds an ACS subdoc and publishes it over rbus via
WIFI_WEBCONFIG_START_ACS (Device.WiFi.X_RDKCENTRAL-COM_StartACS).
The exclude list is taken directly from the channels marked non-operable
(pref == 0) in the Channel Preference TLV, grouped per opclass. No separate
capability merge is needed: the controller already encodes capability
non-operable channels as pref == 0 and the CLI anticipated preferences cannot
override them, so the TLV is the single source of truth.

{
  "Version": "1.0",
  "SubDocName": "Acs",
  "RadioName": "radio1",
  "AcsList": [
    { "opclass": 81, "exclude_channels_length": 2, "exclude_channels": [12,13] }
  ]
}

RadioName is lowercase radio<N>, N = DM index + 1, matches how OneWifi parses it
in convert_radio_name_to_index().

Multi-radio orchestration:
A single CLI POST can target multiple radios. send_channel_sel_request_msg
bundles a Channel Preference + Transmit Power Limit TLV per pending radio into
one 1905 message with a shared msg_id; handle_channel_sel_req splits the
bundled TLVs back per RUID and dispatches one event per radio. The radio
callback iterates all radios in the decoded subdoc and spawns one
em_cmd_op_channel_report_t per radio, tagged with ctx->arr_index so the
orchestrator can match each command to the correct EM. send_channel_sel_response_msg
acknowledges every radio that shares the request msg_id.

Note: Requires corresponding changes in OneWifi for full functionality.

### DIFF
--- a/inc/dm_easy_mesh.h
+++ b/inc/dm_easy_mesh.h
@@ -195,6 +195,14 @@ public:
 	static em_freq_band_t get_freq_band_by_op_class(int op_class);
 
 	/**!
+	 * @brief Returns all op_class numbers for the given frequency band.
+	 *
+	 * @param[in] band Frequency band to filter by.
+	 * @returns Vector of op_class numbers belonging to that band.
+	 */
+	static std::vector<int> get_op_classes_by_band(em_freq_band_t band);
+
+	/**!
 	 * @brief Retrieves the list of channel associated with a given operating class.
 	 *
 	 * This function takes an operating class as input and returns the corresponding
--- a/inc/dm_easy_mesh_agent.h
+++ b/inc/dm_easy_mesh_agent.h
@@ -189,6 +189,27 @@ public:
 	 */
 	int analyze_channel_sel_req(em_bus_event_t *evt, wifi_bus_desc_t *desc,bus_handle_t *bus_hdl);
 
+	/**!
+	 * @brief Builds and sends the ACS exclude-channel list to OneWifi.
+	 *
+	 * Constructs a JSON document containing the channels that ACS must not select,
+	 * derived from the Channel Selection Request TLV and hardware/regulatory capability
+	 * constraints. Publishes the document to OneWifi via the StartACS rbus interface,
+	 * delegating the final channel decision to the platform ACS.
+	 *
+	 * @param[in] desc Pointer to the WiFi bus descriptor to be updated.
+	 * @param[in] bus_hdl Handle to the bus where the event is processed.
+	 * @param[in] channel_sel Pointer to the parsed channel selection request data.
+	 * @param[in] radio_name Radio identifier string (e.g. "radio1") published as RadioName
+	 *                      in the ACS subdoc so OneWifi/vendor ACS can resolve the target radio.
+	 *
+	 * @returns int Status code indicating success or failure of the operation.
+	 * @retval 0 on success.
+	 * @retval Non-zero error code on failure.
+	 */
+	int send_acs_subdoc(wifi_bus_desc_t *desc, bus_handle_t *bus_hdl, op_class_channel_sel *channel_sel,
+	                    const char *radio_name);
+
     /**!
 	 * @brief Analyzes the csa beacon frame received.
 	 *
--- a/inc/em_channel.h
+++ b/inc/em_channel.h
@@ -19,11 +19,13 @@
 #ifndef EM_CHANNEL_H
 #define EM_CHANNEL_H
 
+#include <vector>
 #include "em_base.h"
 
 #define ACK_FROM_AGENT 0
 #define ACK_FROM_CTRL  1
 
+class em_t;
 class em_cmd_t;
 class em_channel_t {
 
@@ -202,7 +204,7 @@ public:
 	 *
 	 * @note Ensure that the buffer is allocated with sufficient size before calling this function.
 	 */
-	short create_channel_pref_tlv(unsigned char *buff);
+	short create_channel_pref_tlv(unsigned char *buff, em_t *target = NULL);
 
 	/**!
 	 * @brief Updates map with non-operable and unsupported channels as per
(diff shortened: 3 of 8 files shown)


nikunjshah-cap:
**Controller behavior**

- Currently, any opclass/channel not present in the data model is treated as non‑operable. Is this intended?
- The controller only adds non‑operable channels to the channel‑selection request, which results in losing existing operable channel preferences from the data model. Can operable channels along with their preference values also be included?

**Agent behavior**

- Agent consolidates operable/non‑operable channels per opclass but does not differentiate primary vs center channel rows as defined in the E4 table.
- It does not account for scenarios where the same channel may be operable in one opclass but inoperable in another (loss of per‑opclass granularity), is there any assumption taken to discount this scenario
- While submitting the ACS sub‑document, only the channel is selected, not the preferred OPCLASS. How is OPCLASS preference intended to be handled?

**Suggestion**

Can we consider excluding all opclass/channels other than the one's with highest preference value to the ACS sub‑document. OPClass/Channel which are not present in the channel selection request can then implicitly be treated with best preference, aligning with standard intent.

dkyncu:
> **Controller behavior**
> 
> * Currently, any opclass/channel not present in the data model is treated as non‑operable. Is this intended?
> * The controller only adds non‑operable channels to the channel‑selection request, which results in losing existing operable channel preferences from the data model. Can operable channels along with their preference values also be included?
> 
> **Agent behavior**
> 
> * Agent consolidates operable/non‑operable channels per opclass but does not differentiate primary vs center channel rows as defined in the E4 table.
> * It does not account for scenarios where the same channel may be operable in one opclass but inoperable in another (loss of per‑opclass granularity), is there any assumption taken to discount this scenario
> * While submitting the ACS sub‑document, only the channel is selected, not the preferred OPCLASS. How is OPCLASS preference intended to be handled?
> 
> **Suggestion**
> 
> Can we consider excluding all opclass/channels other than the one's with highest preference value to the ACS sub‑document. OPClass/Channel which are not present in the channel selection request can then implicitly be treated with best preference, aligning with standard intent.

**Design intent**. In this architecture the EM Controller and Agent do not pick a channel. They only narrow the set of channels the platform's ACS is allowed to use. Final selection is delegated to the vendor ACS through the ACS subdoc. Per FTRS EasyMesh+ [CORE_AIR_0013]: "ACS may not have any restrictions ... other than the static and dynamic non-operable channels" and [CORE_AIR_0014]: "the EasyMesh Agent's channel preference list must reflect the channels that the automatic channel selection daemon can use ... To configure a fixed channel, the Controller preference list will contain only one allowed 20 MHz channel". Width / opclass decisions are explicitly left to the platform ("the platform's channel management / selection shall select a lower channel width").

**Controller only emitting non-operable channels (no operable preference values).** Intentional. IEEE 1905 Channel Preference TLV semantics treat channels not listed in the TLV as implicit best preference, so the encoding is: listed = non-op (0x00), unlisted = operable/best. The platform ACS in this model consumes an allowed set, not a ranking, so intermediate preference values would not be used.

**Opclasses absent from the CLI treated as all non-operable.** Intentional. The primary CLI flow is "user picks a single 20 MHz channel" (e.g., 5 GHz ch 36). To honor that we turn it into an exclude set that isolates the chosen channel: every other channel in the same opclass, plus every channel in the other same-band opclasses, is marked non-operable. That matches FTRS 0014's "one allowed 20 MHz channel -> single-channel lock" directly and prevents ACS from escaping to 40/80/160 MHz bundles that the user did not request. Emitting nothing for a silent opclass would leave it implicitly best and let ACS jump into a wider bundle.

**Primary vs center channel rows in E4.** The agent does not merge rows. Each opclass in m_e4_table is processed independently and the ACS subdoc carries one entry per opclass (AcsList: [{opclass, exclude_channels}, ...]), so per-row granularity is preserved on the wire.

**Same channel operable in one opclass, inoperable in another.** Same channel operable in one opclass, inoperable in another. This case is real (e.g., ch 149 appears in both opclass 124 and 125; a 5 GHz primary also appears under the 80 MHz opclass 128) and is preserved by the per-opclass shape of the subdoc; a physical channel can sit in exclude_channels under one opclass and not under another. No cross-opclass merging happens.

**OPCLASS preference not carried in the ACS subdoc.** By design, per FTRS 0014. Width / opclass choice is the platform ACS's job; Controller and Agent only constrain the allowed channel pool.

**Your suggestion (exclude everything except the highest-preference set; treat absent channels/opclasses as implicit best).** Semantically the controller output already matches that: a single allowed 20 MHz channel surfaces as "everything else excluded"; listed = non-op, unlisted = best, per IEEE 1905. Same on the agent side in the ACS subdoc: exclude_channels explicit, everything else implicitly allowed.

nikunjshah-cap:
Have these changes been validated together with the corresponding OneWiFi/RDK-WIFI-HAL modifications on the BananaPi platform?

If so, could you please share a brief summary of the observed behavior, including the expected vs. actual results?

PR #665, opened 2026-04-27
RDKBWIFI-420: Add implementation for changing the backhaul SSID.

hp490:
Add an implementation in the unified WiFi mesh to update the backhaul SSID via the set SSID method, ensuring the change propagates to all agents in post-order.


src/ctrl/em_ctrl.cpp
```diff
@@ -284,6 +286,84 @@ void em_ctrl_t::handle_set_ssid_list(em_bus_event_t *evt)
...
+ *
+ * @param evt Pointer to the em_bus_event_t containing the event parameters.
+ */
+void em_ctrl_t::handle_set_bh_cfg(em_bus_event_t *evt)
+{
+    em_printfout("Inside handle_set_bh_cfg Function");
+    int ret;
+    dm_easy_mesh_t dm;
+
+    if (m_orch->is_cmd_type_in_progress(evt) == true) {
```
nikunjshah-cap: Should ensure following

-- Once BH Reconfig is initiated, no new commands are accepted from CLI till BH Reconfig is completed
-- In case alternate command is queued, (m_orch->is_cmd_type_in_progress(evt) == true) then should handle the error condition to requeue the command for recursive call and reject initiation in case of first call.

PR #671, opened 2026-05-05
RDKBWIFI-426: Trigger orchestration immediately on command queueing to improve responsiveness

SachinMChougale:
Changes to trigger orchestration event on queuing a command for immediate processing
- Added em_event_type_orch to event enumeration
- Updated em_mgr_t::start() to call the orchestration handler on event
- Updated em_orch_t::submit_command() to trigger orchestration event on queuing command

This ensures commands are processed with minimal latency, improving orchestration responsiveness.

### DIFF
--- a/src/em/em_mgr.cpp
+++ b/src/em/em_mgr.cpp
@@ -548,7 +548,15 @@ int em_mgr_t::start()
                     continue;
                 }
                 pthread_mutex_unlock(&m_queue.lock);
-                if (((evt->type == em_event_type_bus) && ((evt->u.bevt.type == em_bus_event_type_reset) ||
+                if (evt->type == em_event_type_orch) {
+                    if (is_data_model_initialized() == true && started == true) {
+                        // Immediately process orchestration queue for the event
+                        em_orch_t *orch = get_orch();
+                        if (orch != nullptr) {
+                            orch->handle_timeout();
+                        }
+                    }
+                } else if (((evt->type == em_event_type_bus) && ((evt->u.bevt.type == em_bus_event_type_reset) ||
                       (evt->u.bevt.type == em_bus_event_type_get_reset))) ||
 						(is_data_model_initialized() == true)) {
 
--- a/src/orch/em_orch.cpp
+++ b/src/orch/em_orch.cpp
@@ -141,6 +141,15 @@ bool em_orch_t::submit_command(em_cmd_t *pcmd)
     } else {
         queue_push(m_pending, pcmd);
         push_stats(pcmd);
+        // trigger orchestration event to process the command
+        em_event_t *evt = static_cast<em_event_t *>(malloc(sizeof(em_event_t)));
+        if (evt == nullptr) {
+            em_printfout("Failed to allocate memory for orchestration event");
+            // Command is queued for processing, count not submit event
+        } else {
+            evt->type = em_event_type_orch;
+            m_mgr->push_to_queue(evt);
+        }
         submitted = true;
     }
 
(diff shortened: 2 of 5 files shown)


vlad-safonov:
I already investigated similar failures today and just wanted to share - build check fails because github updated ubuntu:latest and now it is ubunut:26.04 (with OpenSSL 3.5) but before it was based on ubuntu24.04 (with OpenSSL 3.0). If change in .github/workflows/build.yml 
```
- container: 'ubuntu:latest'
+ container: 'ubuntu:24.04' #'ubuntu:latest'
```
it will be fine again

src/em/em_mgr.cpp
```diff
@@ -548,7 +548,15 @@ int em_mgr_t::start()
                     continue;
                 }
                 pthread_mutex_unlock(&m_queue.lock);
-                if (((evt->type == em_event_type_bus) && ((evt->u.bevt.type == em_bus_event_type_reset) ||
+                if (evt->type == em_event_type_orch) {
+                    if (is_data_model_initialized() == true && started == true) {
```
Nikita-Hakai, line 552: This implies to both Controller and agent right? I see that is_data_model_initialized() is always true in the agent. Do you see any implications on that?

src/orch/em_orch.cpp
```diff
@@ -141,6 +141,16 @@ bool em_orch_t::submit_command(em_cmd_t *pcmd)
     } else {
         queue_push(m_pending, pcmd);
         push_stats(pcmd);
+        // trigger orchestration event to process the command
+        em_event_t *evt = static_cast<em_event_t *>(malloc(sizeof(em_event_t)));
```
Nikita-Hakai, line 145: Won't there a new em_event_type_orch event be generated for every command submitted in this case, is that the intent?

src/em/em_mgr.cpp
```diff
@@ -548,7 +548,15 @@ int em_mgr_t::start()
...
+                        // Immediately process orchestration queue for the event
+                        em_orch_t *orch = get_orch();
+                        if (orch != nullptr) {
+                            orch->handle_timeout();
```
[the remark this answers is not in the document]
SachinMChougale: I think this is unlikely to be a real regression. The orchestration queue  and start_complete() was already being handled from the timeout path before. This change just have the handing of orchestration queue queue quiuckly only if start_complete() was done before.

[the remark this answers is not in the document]
SachinMChougale: This wakeup path is just for handling orchestration queue only if data model is initialised and started == true.
If any of these conditions failed, then the orchestration queue gets handled automatically in timeout path.

src/em/em_mgr.cpp
SachinMChougale: I dont think there will be any issue of  is_data_model_initialized() being always true. On the Agent, orchestration events can still be blocked until started == true is flipped by the first timeout.

src/orch/em_orch.cpp
SachinMChougale: Yes — that is the intent. By posting an event per submitted command, we ensure low-latency processing of orchestration work.

1. submit_command() pushes the command into m_pending. It then creates an em_event_type_orch events and sends it to m_mgr->push_to_queue().
2. That event is the wake-up trigger telling the manager: “there is new orchestration work pending.”

This to make sure that orchestration commands gets processed quickly, rather than waiting for next timeout to process it.

Nikita-Hakai (approved):

PR #688, opened 2026-05-28
RDKBWIFI-451: EasyMesh controller failing on configuration due to failed validation of mandatory TLVs

aprasad-97:
Reason for change: Enabled validation on AP CAP response msg. This adds additional protective measures on TLVs inserted in response.
Risks: Low


src/em/capability/em_capability.cpp
vlad-safonov: If you want to enable validation need also uncomment `errors` definition

Nikita-Hakai (approved):

PR #681, opened 2026-05-21
RDKBWIFI-446 Controller initiated Steering using Rbus Method

anandhkr-plume:
Implemented code changes needed for controller initiated steering. Includes Steering Mandate and Steering Opportunity for both 11v and non 11v Clients.

Test Procedure: Trigger Client Steering using Rbus Method, and check Client is steered across Band/Device
    rbuscli method_values Device.WiFi.DataElements.Network.Device.{i}.Radio.{i}.BSS.{i}.STA.{i}.ClientSteer

Signed-off-by: Amalesh Nandh K R <anandhkr@plume.com>

PR #692, opened 2026-05-29
RDKBWIFI-278: Refactor CACR add multi-STA support and received subdoc from UWM

sundram0711:
- Added rdkb_cli API support to trigger Client Association Control Request (CACR)
- Refactored implementation to support multiple STA entries per message (spec compliant)

### DIFF
--- a/inc/em_base.h
+++ b/inc/em_base.h
@@ -106,6 +106,7 @@ extern "C"
 #define EM_MAX_CLIENT_ASSOC_CTRL_REQ_TX_THRESH  5
 #define MAX_STA_TO_DISASSOC		32
 #define EM_MAX_DB_CFG_CRITERIA	32
+#define MAX_STA_LIST 20
 
 #define EM_CLI_MAX_ARGS 5
 
@@ -365,6 +366,10 @@ static const mac_address_t EM_GLOBAL_MAC_ADDRESS = {0xff, 0xff, 0xff, 0xff, 0xff
 #define EM_HE_PHY4_MU_BEAMFORMER     0x02
 #define EM_HE_PHY4_BFEE_STS_LE80_MASK  (0x07 << 2) /* beamformee STS <= 80 MHz */
 #define EM_HE_PHY4_BFEE_STS_GT80_MASK  (0x07 << 5) /* beamformee STS > 80 MHz */
+
+//CACR assoc control values
+#define ASSOC_CONTROL_BLOCK 0x00
+#define ASSOC_CONTROL_UNBLOCK 0x01
 
 typedef char em_interface_name_t[32];
 typedef unsigned char em_nonce_t[16];
@@ -1098,8 +1103,8 @@ typedef struct {
     bssid_t 	bssid;
     unsigned char assoc_control;
     unsigned short validity_period;
-    unsigned char count;
-    mac_address_t sta_mac;
+    unsigned char sta_count;
+    mac_address_t sta_list[MAX_STA_LIST];
 }__attribute__((__packed__)) em_client_assoc_ctrl_req_t;
 
 typedef struct {
@@ -2339,6 +2344,7 @@ typedef enum {
     em_state_ctrl_bsta_cap_pending,
     em_state_ctrl_topo_publish_pending,
     em_state_ctrl_unassoc_sta_link_metrics_pending,
+    em_state_ctrl_client_assoc_ctrl_req_pending,
 
     //common states
     em_state_beacon_report_pending,
@@ -2397,6 +2403,7 @@ typedef enum {
     em_cmd_type_get_link_quality_report,
     em_cmd_type_unassoc_sta_query,
     em_cmd_type_unassoc_sta_result,
+    em_cmd_type_client_assoc_ctrl_req,
 
     em_cmd_type_max,
 } em_cmd_type_t;
@@ -3195,6 +3202,7 @@ typedef enum {
     dm_orch_type_link_quality_report,
     dm_orch_type_unassoc_sta_link_req_query,
     dm_orch_type_unassoc_sta_result,
+    dm_orch_type_client_assoc
 
 } dm_orch_type_t;
 
@@ -3320,6 +3328,14 @@ typedef struct {
 } em_cmd_btm_report_params_t;
 
 typedef struct {
+    bssid_t 	bssid;
+    unsigned char assoc_control;
+    unsigned short validity_period;
+    unsigned char sta_count;
+    mac_address_t sta_list[MAX_STA_LIST];
+}em_cmd_client_assoc_params_t;
+
+typedef struct {
     mac_address_t	sta_mac;
     bssid_t	bssid;
     unsigned int disassoc_time;
@@ -3384,6 +3400,7 @@ typedef struct {
         em_cmd_steer_params_t	steer_params;
         em_cmd_btm_report_params_t  btm_report_params;
         em_cmd_disassoc_params_t	disassoc_params;
+        em_cmd_client_assoc_params_t   client_assoc_params;
 		em_cmd_scan_params_t	scan_params;
         em_cmd_ap_metrics_rprt_params_t ap_metrics_params;
         em_cmd_unassoc_sta_query_params_t unassoc_sta_query_params;
--- a/inc/em_cmd_client_assoc_ctrl_req.h
+++ b/inc/em_cmd_client_assoc_ctrl_req.h
@@ -1,0 +1,21 @@
+#ifndef EM_CMD_CLIENT_ASSOC_CTRL_REQ_H
+#define EM_CMD_CLIENT_ASSOC_CTRL_REQ_H
+
+#include "em_cmd.h"
+
+class em_cmd_client_assoc_ctrl_req_t: public em_cmd_t {
+
+public:
+
+    /**!
+     * @brief Constructs a client assoc ctrl req command.
+     *
+     * @param[in] params The client assoc ctrl req parameters.
+     * @param[in] dm     dm reference.
+     * @param[in] svc    Service type (controller or agent). Defaults to controller.
+     */
+    em_cmd_client_assoc_ctrl_req_t(em_cmd_client_assoc_params_t params, dm_easy_mesh_t& dm,
+        em_service_type_t svc = em_service_type_ctrl);
+};
+
+#endif
--- a/src/cmd/em_cmd_client_assoc_ctrl_req.cpp
+++ b/src/cmd/em_cmd_client_assoc_ctrl_req.cpp
@@ -1,0 +1,46 @@
+/**
+ * Copyright 2023 Comcast Cable Communications Management, LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include "em_cmd_client_assoc_ctrl_req.h"
+
+em_cmd_client_assoc_ctrl_req_t::em_cmd_client_assoc_ctrl_req_t(em_cmd_client_assoc_params_t params,
+    dm_easy_mesh_t& dm, em_service_type_t svc)
+{
+    em_cmd_ctx_t ctx;
+
+    m_type = em_cmd_type_client_assoc_ctrl_req;
+    memcpy(&m_param.u.client_assoc_params, &params, sizeof(em_cmd_client_assoc_params_t));
+
+    memset(reinterpret_cast<unsigned char *>(&m_orch_desc[0]), 0, EM_MAX_CMD * sizeof(em_orch_desc_t));
+
+    m_orch_op_idx = 0;
+    m_num_orch_desc = 1;
+    m_orch_desc[0].op = dm_orch_type_client_assoc;
+    m_orch_desc[0].submit = true;
+
+    strncpy(m_name, "client_assoc", strlen("client_assoc") + 1);
+    m_svc = svc;
+    init(dm);
+
+    memset(&ctx, 0, sizeof(em_cmd_ctx_t));
+    ctx.type = m_orch_desc[0].op;
+    m_data_model.set_cmd_ctx(&ctx);
+}
--- a/src/ctrl/dm_easy_mesh_ctrl.cpp
+++ b/src/ctrl/dm_easy_mesh_ctrl.cpp
@@ -64,6 +64,7 @@
 #include "em_cmd_mld_reconfig.h"
 #include "em_cmd_bsta_cap.h"
 #include "em_cmd_unassoc_sta_query.h"
+#include "em_cmd_client_assoc_ctrl_req.h"
 
 extern em_network_topo_t *g_network_topology;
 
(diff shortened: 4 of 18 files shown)


sundram0711:
Hi,
This PR is not ready for review, currently it's in draft stage.

src/em/steering/em_steering.cpp
```diff
@@ -50,33 +51,38 @@ int em_steering_t::send_client_assoc_ctrl_req_msg()
...
-                memset(&assoc_ctrl, 0, sizeof(assoc_ctrl));
+    for (unsigned int j = 0; j < dm->m_num_bss; j++) {
+        if ((memcmp(assoc_param->bssid, dm->m_bss[j].m_bss_info.bssid.mac, sizeof(mac_address_t)) == 0) &&
+         (memcmp(dm->m_bss[j].m_bss_info.ruid.mac, get_radio_interface_mac(), sizeof(mac_address_t)) == 0)) {
 
-                memcpy(&assoc_ctrl.bssid, &disassoc_param->bssid, sizeof(bssid_t));
-                //Current Implementation for One Sta Per Message
-                assoc_ctrl.count = 1;
-                memcpy(&assoc_ctrl.sta_mac, &disassoc_param->sta_mac, sizeof(mac_address_t));
+            em_client_assoc_ctrl_req_t assoc_ctrl;
```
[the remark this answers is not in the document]
sundram0711: invalid bssid check is already handled at orchestrator level

src/orch/em_orch_ctrl.cpp
```diff
@@ -701,6 +709,19 @@ unsigned int em_orch_ctrl_t::build_candidates(em_cmd_t *pcmd)
                 }
                 break;
 
+            case em_cmd_type_client_assoc_ctrl_req:
+                if (em->is_al_interface_em() == false) {
+                    em_cmd_client_assoc_params_t *assoc = &pcmd->m_param.u.client_assoc_params;
+
+                    if (em->find_bss(assoc->bssid) == NULL) {
+                        em_printfout("%s:%d: Skipping radio (BSSID not owned by this EM)\n", __func__, __LINE__);
```
nikunjshah-cap, line 717: __func__, __LINE__, and the trailing newline are already handled by em_printfout, making them redundant in the caller. Please remove these from all em_printfout usages to keep the logging consistent and avoid duplication

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -2539,6 +2540,142 @@ int dm_easy_mesh_ctrl_t::analyze_command_steer(em_bus_event_t *evt, em_cmd_t *cm
...
+        cJSON_Delete(obj);
+        return 0;
+    }
+    for (int i = 0; i < 6; i++) {
+        assoc_param.bssid[i] = static_cast<unsigned char>(mac[i]);
+    }
+
+    if ((assoc_control_obj = cJSON_GetObjectItem(wfa_obj, "AssocControl")) != NULL) {
+        double v = cJSON_GetNumberValue(assoc_control_obj);
+        if (v != 0.0 && v != 1.0) {
```
nikunjshah-cap, line 2597: As per specification, Association control can have following values:

0x00: Block
0x01: Unblock
0x02: Timed block
0x03: Indefinite block

Currently RDK Agent only supports Block and Unblock, it would be beneficial for the Controller to support all defined values to improve interoperability.

Additionally, better to use the corresponding ASSOC_CONTROL macros instead of hardcoded values to improve readability and maintainability

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -2539,6 +2540,142 @@ int dm_easy_mesh_ctrl_t::analyze_command_steer(em_bus_event_t *evt, em_cmd_t *cm
...
+            cJSON_Delete(obj);
+            return 0;
+        }
+        for (int k = 0; k < 6; k++) {
+            assoc_param.sta_list[valid_sta_count][k] = static_cast<unsigned char>(mac[k]);
+        }
+        valid_sta_count++;
+    }
+
+    if (valid_sta_count == 0) {
```
nikunjshah-cap, line 2658: In the invalid-entry case, the function returns immediately. In the success path, valid_sta_count will always be equal to sta_count, so maintaining two separate variables does not seem necessary

src/cmd/em_cmd_client_assoc_ctrl_req.cpp
```diff
@@ -0,0 +1,28 @@
+#include <stdio.h>
```
nikunjshah-cap, line 1: Add the copyright header aligned with other em_cmd_* files

src/em/steering/em_steering.cpp
```diff
@@ -50,33 +51,38 @@ int em_steering_t::send_client_assoc_ctrl_req_msg()
...
-        for (unsigned int j = 0; j < dm->m_num_bss; j++) {
-            if ((memcmp(disassoc_param->bssid, dm->m_bss[j].m_bss_info.bssid.mac, sizeof(bssid_t)) == 0) &&
-                (memcmp(dm->m_bss[j].m_bss_info.ruid.mac, get_radio_interface_mac(), sizeof(mac_address_t)) == 0)) {
+    if (assoc_param->sta_count == 0 || assoc_param->sta_count > MAX_STA_LIST) {
+        em_printfout("%s:%d ERROR: Invalid sta_count=%u (max %u)", __func__, __LINE__, assoc_param->sta_count, MAX_STA_LIST);
+        set_state(em_state_ctrl_configured);
+        return -1;
+    }
+
+    if (assoc_param->assoc_control != 0x00 && assoc_param->assoc_control != 0x01) {
```
nikunjshah-cap, line 62: Update to handle following values as per specification:

0x00: Block
0x01: Unblock
0x02: Timed block
0x03: Indefinite block

src/em/steering/em_steering.cpp
```diff
@@ -528,20 +620,57 @@ int em_steering_t::handle_client_assoc_ctrl_req(unsigned char *buff, unsigned in
...
+        em_printfout("%s:%d: invalid ClientAssocCtrlRequest TLV (sta_count=%u tlv_len=%u expected_min=%zu max=%u)",
+            __func__, __LINE__, sta_count, assoc_tlv_len, expected_len, MAX_STA_LIST);
+        return -1;
     }
 
-    send_1905_ack_message(assoc_ctrl_req->sta_mac, msg_id, 0, hdr->src);
+    // Collect indices of associated and unassociated STAs
+    std::vector<uint8_t> associated_sta_indices;
+
+    if (assoc_ctrl_req->assoc_control == 0x00) {
```
nikunjshah-cap, line 640: Better to define and use an ASSOC_CONTROL macro instead of an absolute value

src/em/steering/em_steering.cpp
nikunjshah-cap, line 89: By transitioning directly to em_state_ctrl_configured, the Controller orch state machine does not wait for the Agent ACK as required by the specification. As a result, in the event of a transmission failure it will not retry the message.

Instead, the state machine should transition to an intermediate pending state after sending the request. Upon receipt of the ACK, it can then move to em_state_ctrl_configured and complete the orchestration. 

This ensures compliance with the expected ACK handling and retry behavior.

nikunjshah-cap (dismissed):

src/ctrl/dm_easy_mesh_ctrl.cpp
sundram0711: Thanks for the suggestion.

The Timed Block (0x02) and Indefinite Block (0x03) Association Control values were introduced in later EasyMesh revisions. Since the current work is focused on the existing Association Control functionality, I'd prefer to keep the scope of this PR limited and address support for those additional modes in a follow-up enhancement.

I do agree with replacing the hardcoded values with the corresponding ASSOC_CONTROL macros and will update that as part of this PR.

src/em/steering/em_steering.cpp
sundram0711: Thanks for the suggestion.

The Timed Block (0x02) and Indefinite Block (0x03) Association Control values were introduced in later EasyMesh revisions. Since the current work is focused on the existing Association Control functionality, I'd prefer to keep the scope of this PR limited and address support for those additional modes in a follow-up enhancement.

inc/em_cmd_client_assoc_ctrl_req.h
```diff
@@ -0,0 +1,21 @@
...
+     *
+     * @param[in] params The client assoc ctrl req parameters.
+     * @param[in] dm     dm reference.
+     * @param[in] svc    Service type (controller or agent). Defaults to controller.
+     */
+    em_cmd_client_assoc_ctrl_req_t(em_cmd_client_assoc_params_t params, dm_easy_mesh_t& dm,
+        em_service_type_t svc = em_service_type_ctrl);
+};
+
+#endif
```
Nikita-Hakai, line 21: Can you re-use the em_cmd_sta_steer_t() and use override_op with op = dm_orch_type_client_assoc

src/cmd/em_cmd_client_assoc_ctrl_req.cpp
```diff
@@ -0,0 +1,46 @@
...
+    m_orch_desc[0].submit = true;
+
+    strncpy(m_name, "client_assoc", strlen("client_assoc") + 1);
+    m_svc = svc;
+    init(dm);
+
+    memset(&ctx, 0, sizeof(em_cmd_ctx_t));
+    ctx.type = m_orch_desc[0].op;
+    m_data_model.set_cmd_ctx(&ctx);
+}
```
Nikita-Hakai, line 46: Can you re-use the em_cmd_sta_steer_t() and use override_op with op = dm_orch_type_client_assoc

inc/em_base.h
```diff
@@ -1097,8 +1102,8 @@ typedef struct {
     bssid_t 	bssid;
     unsigned char assoc_control;
     unsigned short validity_period;
-    unsigned char count;
-    mac_address_t sta_mac;
+    unsigned char sta_count;
+    mac_address_t sta_list[MAX_STA_LIST];
```
Nikita-Hakai, line 1107: Do you think we can make this list dynamic?

inc/em_base.h
```diff
@@ -3314,6 +3322,14 @@ typedef struct {
     bssid_t	target;
 } em_cmd_btm_report_params_t;
 
+typedef struct {
+    bssid_t 	bssid;
+    unsigned char assoc_control;
+    unsigned short validity_period;
+    unsigned char sta_count;
+    mac_address_t sta_list[MAX_STA_LIST];
```
Nikita-Hakai, line 3335: Do you think we can make this list dynamic?

src/em/steering/em_steering.cpp
```diff
@@ -155,7 +157,8 @@ int em_steering_t::send_client_assoc_ctrl_req_msg(em_client_assoc_ctrl_req_t *as
     }
 
     m_client_assoc_ctrl_req_tx_cnt++;
-    printf("%s:%d: Client Assoc Control Request (%d) Send Successful\n", __func__, __LINE__, m_client_assoc_ctrl_req_tx_cnt);
+    m_client_assoc_ctrl_req_msg_id= ntohs(cmdu->id);
+    printf("%s:%d: Client Assoc Control Request (%d) Send Successful, msg_id:\n", __func__, __LINE__, m_client_assoc_ctrl_req_tx_cnt, m_client_assoc_ctrl_req_msg_id);
```
Nikita-Hakai: to be replaced with em_printfout

src/cmd/em_cmd_client_assoc_ctrl_req.cpp
sundram0711: Hi @Nikita-Hakai,
could you please clarify your suggestion? em_cmd_sta_steer_t carries em_cmd_steer_params_t, while CACR uses em_cmd_client_assoc_params_t. Do you mean reusing only the orchestration pattern with override_op(dm_orch_type_client_assoc), or reusing the steer command class "em_cmd_sta_steer_t()" itself?

inc/em_base.h
sundram0711: Do you mean replacing sta_list[MAX_STA_LIST] with std::vector<mac_address_t> in the command/request structures?

src/cmd/em_cmd_client_assoc_ctrl_req.cpp
Nikita-Hakai: Okay, in that case maybe you could update em_cmd_sta_steer_t to take generic params like em_cmd_client_cap_report_t or em_cmd_channel_pref_query_t etc?

sundram0711: Thanks Nikita. Just to confirm the expected approach, are you suggesting refactoring em_cmd_sta_steer_t into a generic single-operation command that takes em_cmd_params_t, em_cmd_type_t, dm_orch_type_t, and command name (e.g. "steer_sta" / "client_assoc") as inputs, and then reusing it for both STA Steering and CACR?

Also, would you still expect the orch op to be set using override_op(), or should it be passed directly during command construction?

Do you have any reference implementation in the codebase that already follows this generic-command approach that I can use as a model?

One additional consideration: em_cmd_sta_steer_t is currently used by the Plume's Client Steering work (PR #681). If I refactor em_cmd_sta_steer_t into a generic command, that PR may also need updates to align with the new constructor/interface. Just wanted to confirm this is the preferred direction before making the changes.

sundram0711: Hi @Nikita-Hakai,
If the goal is to avoid creating dedicated .h/.cpp files for single-operation commands, we could consider a follow-up JIRA to introduce a generic command constructor that accepts parameters such as em_cmd_params_t, em_cmd_type_t, dm_orch_type_t, and command name (e.g. "steer_sta" / "client_assoc"), and reuse it across similar command implementations.

For this PR, I followed the existing architecture and implementation pattern already used in other features where dedicated command classes were introduced, such as Backhaul Steering (PR-636) and Unassociated STA Link Metrics (PR-684). Given that em_cmd_sta_steer_t is also currently used by the Plume Client Steering work (PR-681).

What do you think? Does this make sense, or would you still prefer that I refactor em_cmd_sta_steer_t as part of this PR?

inc/em_base.h
sundram0711: Currently the TLV payload size is already dynamic based on sta_count, while the internal storage uses a fixed-size array. Converting to std::vector would require updates across the command, orchestration, serialization, and bus-event handling paths.

src/cmd/em_cmd_client_assoc_ctrl_req.cpp
Nikita-Hakai: I was considering we can optimize the creation of new cmd's for each new operation and instead group similar ones. But if that goes out of scope of this PR, we can take it up separately or please do consider when in future you have similar contributions to make. Agreed that we can take this up as a follow-up JIRA.

sundram0711: Yeah sure, thanks

inc/em_cmd_client_assoc_ctrl_req.h
sundram0711: As discussed, we can follow up these changes in separate PR.

src/validation_test.cpp
```diff
@@ -1264,29 +1264,41 @@ int em_testValidation_t::test_bh_steering_req(unsigned char *buff,int i,unsigned
...
     }
 
     // End of message
-    tlv = (em_tlv_t *)tmp;
+    em_tlv_t *tlv = reinterpret_cast<em_tlv_t *>(tmp);
     tlv->type = em_tlv_type_eom;
     tlv->len = 0;
 
-    tmp += (sizeof (em_tlv_t));
     len += (sizeof (em_tlv_t));
```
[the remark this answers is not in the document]
sundram0711: not critical

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -3063,6 +3064,135 @@ int dm_easy_mesh_ctrl_t::analyze_command_steer(em_bus_event_t *evt, em_cmd_t *cm
...
+    }
+
+    int sta_count = cJSON_GetArraySize(sta_list_obj);
+    if (sta_count <= 0) {
+        em_printfout("%s:%d: StaMacList cannot be empty", __func__, __LINE__);
+        cJSON_Delete(obj);
+        return 0;
+    }
+    if (sta_count > MAX_STA_LIST) {
+        em_printfout("%s:%d: StaMacList too large (%d), max allowed=%d", __func__, __LINE__, sta_count, MAX_STA_LIST);
```
hp490: There’s no need to explicitly print the function name and line number; em_printfout will handle that automatically.

src/cmd/em_cmd_client_assoc_ctrl_req.cpp
```diff
@@ -0,0 +1,46 @@
...
+    memset(reinterpret_cast<unsigned char *>(&m_orch_desc[0]), 0, EM_MAX_CMD * sizeof(em_orch_desc_t));
+
+    m_orch_op_idx = 0;
+    m_num_orch_desc = 1;
+    m_orch_desc[0].op = dm_orch_type_client_assoc;
+    m_orch_desc[0].submit = true;
+
+    strncpy(m_name, "client_assoc", strlen("client_assoc") + 1);
+    m_svc = svc;
+    init(dm);
```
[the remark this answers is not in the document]
sundram0711: No sure about these changes, I followed the existing cmd constructor code

src/em/em.cpp
```diff
@@ -365,14 +369,13 @@ void em_t::proto_process(unsigned char *data, unsigned int len)
...
                 em_configuration_t::process_msg(data, len);
-            } else if (m_sm.get_state() == em_state_ctrl_sta_steer_pending) {
-                em_steering_t::process_msg(data, len);
             } else if (m_sm.get_state() == em_state_ctrl_unassoc_sta_link_metrics_pending) {
                 em_metrics_t::process_msg(data, len);		    
             } else {
                 em_policy_cfg_t::process_msg(data, len);
                 em_channel_t::process_msg(data, len);
                 em_metrics_t::process_msg(data, len);
+                em_steering_t::process_msg(data, len);
```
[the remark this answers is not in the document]
sundram0711: Yeah understand, but I was not having any other option, for client assoc control req implementation, I'm handling the ack in the em_steering.cpp file where I'm validating the radios and taking the correct em, to find the correct the controller state I need to get the right em, because at this point if I trigger cacr for 5GHz, then the ack is coming for for the 2.4Gz radios which is not in the cacr pending state, so this validation I'm doing it in the handle ack message there I'm taking the em instance and reading the right state

Nikita-Hakai (approved):

PR #693, opened 2026-06-01
RDKBWIFI-458: Adding DFS channels to channel preference report message

akshaya3495:
- Added a DFS opclass→channel lookup table plus helpers to add/remove DFS channels from channel preference entries in the data model.
- Updated Channel Preference TLV (agent) generation to only include preference entries marked pref_valid.
- Updated Channel Preference Report message creation to apply DFS/non-DFS adjustments for 5 GHz radios before TLV generation.

### DIFF
--- a/src/em/channel/em_channel.cpp
+++ b/src/em/channel/em_channel.cpp
@@ -51,6 +51,172 @@
  */
 constexpr unsigned char BSS_COLOR_BSS_LOAD_PRESENT = 0x80;
 
+typedef struct {
+    unsigned int op_class;
+    unsigned int num_channels;
+    unsigned char channels[EM_MAX_CHANNELS_IN_LIST];
+} dfs_op_class_entry;
+
+constexpr dfs_op_class_entry dfs_table[] = {
+    {118, 4,  {52, 56, 60, 64}},
+    {119, 2,  {52, 60}},
+    {120, 2,  {56, 64}},
+    {121, 12, {100,104,108,112,116,120,124,128,132,136,140,144}},
+    {122, 6,  {100,108,116,124,132,140}},
+    {123, 6,  {104,112,120,128,136,144}},
+    {128, 4,  {58,106,122,138}},
+    {129, 2,  {50,114}},
+    {130, 4,  {58,106,122,138}}
+};
+
+constexpr unsigned int dfs_table_size = sizeof(dfs_table) / sizeof(dfs_table[0]);
+
+bool em_channel_t::is_dfs_channel(unsigned int op_class, unsigned int channel)
+{
+    for (unsigned int i = 0; i < dfs_table_size; ++i) {
+        if (dfs_table[i].op_class != op_class) {
+            continue;
+        }
+
+        for (unsigned int j = 0; j < dfs_table[i].num_channels; ++j) {
+            if (dfs_table[i].channels[j] == channel) {
+                return true;
+            }
+        }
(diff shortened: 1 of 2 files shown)


src/em/channel/em_channel.cpp
```diff
@@ -51,6 +51,171 @@
...
+    em_device_info_t *device = dm->get_device_info();
+    if (!device) {
+        return -1;
+    }
+
+    unsigned int num_opclass = dm->get_num_op_class();
+    if (device->dfs_enable) {
+	short rc = remove_dfs_channels_from_noop_pref_in_dm(dm, ruid);
+	return (rc < 0) ? rc : static_cast<short>(num_opclass);
+    }
```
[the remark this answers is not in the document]
akshaya3495: The DFS channels are added to the Channel Preference Report as non-operable channels. Therefore, when dfs_enable is set, DFS channels marked with Non-operable preference are removed from the Channel Preference Report.

src/em/channel/em_channel.cpp
```diff
@@ -51,6 +51,172 @@
...
+            *op = dm_op_class_t{};
+
+            em_op_class_info_t *info = &op->m_op_class_info;
+            memcpy(info->id.ruid, ruid, sizeof(mac_address_t));
+            info->id.type = em_op_class_type_preference;
+            info->id.op_class = entry.op_class;
+            info->op_class = entry.op_class;
+
+	    pref_map[entry.op_class] = op;
+	    ++num_opclass;
```
Nikita-Hakai: indentation mismatch

src/em/channel/em_channel.cpp
```diff
@@ -51,6 +51,172 @@
  */
 constexpr unsigned char BSS_COLOR_BSS_LOAD_PRESENT = 0x80;
 
+typedef struct {
+    unsigned int op_class;
+    unsigned int num_channels;
+    unsigned char channels[EM_MAX_CHANNELS_IN_LIST];
+} dfs_op_class_entry;
+
+constexpr dfs_op_class_entry dfs_table[] = {
```
Nikita-Hakai, line 60: Instead of a new table, can we re-use `dm_easy_mesh_t::m_e4_table[]` maybe with an is_dfs flag per entry?

src/em/channel/em_channel.cpp
```diff
@@ -51,6 +51,172 @@
...
+    }
+
+    // Add the new channel at the next available index
+    unsigned int idx = info->num_channels++;
+    info->channels[idx] = channel;
+    info->channel_pref[idx] = 0x00;
+    return true;
+}
+
+short em_channel_t::check_dfs_flag_and_update_noop_channels(const mac_address_t ruid)
```
Nikita-Hakai, line 145: This function seems to be directly updating the data model while constructing the pref report msg, is this intentional? I think if this is intentional, it should be done before the cmd reaches msg creation stage.

src/em/channel/em_channel.cpp
```diff
@@ -51,6 +51,172 @@
...
+    for (unsigned int i = 0; i < info->num_channels; ++i) {
+        if (info->channels[i] == channel) {
+            return false;
+        }
+    }
+
+    // Add the new channel at the next available index
+    unsigned int idx = info->num_channels++;
+    info->channels[idx] = channel;
+    info->channel_pref[idx] = 0x00;
```
Nikita-Hakai, line 141: What is the need for this hardcoded value of 0x00 here?

PR #697, opened 2026-06-09
Reverting orchestration changes done for fast onboarding -  which are causing some raceconditions 

SachinMChougale:
Note: Changes will be added back later when these race conditions are fixed


Nikita-Hakai (approved):

nikunjshah-cap (commented):
This PR is to revert the update to trigger the event on state change. Need to be added with updates to the SM avoiding race conditions for stability

PR #714, opened 2026-06-25
[RDK Controller/Agent] DFS Events Handling

prathbr:
Agent
- Processing the DFSEvent JSON received from oneWifi
- Updating DataBase and DataModel with DFS Status for available/non occ/active on RDK Agent
- Sending Unsolicted Channel Preference Report for any DFS State change Controller
- Processing Unsolicited Channel Preference Report received by Agent
- Storing the DFS status from CAC Status Report TLV and CAC Completion Report TLVs into DataModel/DataBase
- Excluding non occupancy channels from Channel Preference TLV in Channel Selection Request

### DIFF
--- a/src/agent/dm_easy_mesh_agent.cpp
+++ b/src/agent/dm_easy_mesh_agent.cpp
@@ -386,7 +386,7 @@ int dm_easy_mesh_agent_t::analyze_onewifi_vap_cb(em_bus_event_t *evt, em_cmd_t *
     return num;
 }
 
-int dm_easy_mesh_agent_t::analyze_onewifi_radio_cb(em_bus_event_t *evt, em_cmd_t *pcmd[])
+int dm_easy_mesh_agent_t::analyze_onewifi_radio_cb(em_bus_event_t *evt, em_cmd_t *pcmd[], em_mgr_t *mgr)
 {
     webconfig_t config;
     webconfig_external_easymesh_t ext;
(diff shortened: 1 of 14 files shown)


src/agent/dm_easy_mesh_agent.cpp
```diff
@@ -661,6 +703,175 @@ int dm_easy_mesh_agent_t::analyze_channel_sel_req(em_bus_event_t *evt, wifi_bus_
 
 }
 
+int dm_easy_mesh_agent_t::analyze_dfs_state(em_bus_event_type_dfs_event_params_t *dfs_params,em_mgr_t *mgr)
+{
+    dm_easy_mesh_t *dm;
+    dm = mgr->get_data_model(NULL, NULL);
+    em_channel_pref_reason_t dfs_event_reason;
```
SachinMChougale, line 710: dfs_event_reason is uninitialized for em_dfs_event_type_started, but is passed to send_unsolicited_channel_pref_report()

src/em/channel/em_channel.cpp
```diff
@@ -60,7 +60,7 @@ short em_channel_t::create_channel_pref_tlv_agent(unsigned char *buff, unsigned
     dm_easy_mesh_t *dm;
     dm_op_class_t	*op_class;
     unsigned char *tmp;
-    unsigned char pref_bits = EM_CH_PREF_NON_OPERABLE | em_channel_pref_reason_t::em_channel_pref_reason_operation_disallowed_regulatory_restriction;
+    unsigned char pref_bits = EM_CH_PREF_NON_OPERABLE | m_current_reason;
```
SachinMChougale, line 63: m_current_reason - was never reset and we may endup using previously stored reason for synchronous channel preference report. Better pass the reason explicitly per report

src/agent/dm_easy_mesh_agent.cpp
```diff
@@ -418,19 +418,61 @@ int dm_easy_mesh_agent_t::analyze_onewifi_radio_cb(em_bus_event_t *evt, em_cmd_t
         em_printfout("Radio subdoc decode fail");
```
SachinMChougale, line 418: Suggestion: Return 0 on decode failure

src/agent/dm_easy_mesh_agent.cpp
```diff
@@ -418,19 +418,61 @@ int dm_easy_mesh_agent_t::analyze_onewifi_radio_cb(em_bus_event_t *evt, em_cmd_t
...
+    em_radio_info_t *radio_info = dm.get_radio(index)->get_radio_info();
+    dm_easy_mesh_t::macbytes_to_string(radio_info->intf.mac, mac_str);
+    cm_config.type = em_commit_target_radio;
+    snprintf(reinterpret_cast<char *> (cm_config.params), sizeof(cm_config.params), "%s", mac_str);
+
+    /*
+     * Analyze DFS state for channels in dm that are in CAC Non-Occupancy state
+     * when a matching DFS event is found.
+     */
+    for (unsigned int oc = 0; oc < m_num_opclass; oc++) {
```
SachinMChougale, line 430: function scans all op classes/events twice: approximately O(opclasses * events) twice. Build a single event lookup keyed by (op_class, channel, radio_index), then process only matching CAC/current entries.

src/agent/dm_easy_mesh_agent.cpp
```diff
@@ -418,19 +418,61 @@ int dm_easy_mesh_agent_t::analyze_onewifi_radio_cb(em_bus_event_t *evt, em_cmd_t
...
+            }
+        }
+    }
 
-	while ((pcmd[num] = tmp->clone_for_next()) != NULL) {
-		tmp = pcmd[num];
-		num++;
-	}
-	return num;
+    commit_config(dm, cm_config);
```
SachinMChougale, line 443: do commit_config before analyze_dfs_state

src/agent/dm_easy_mesh_agent.cpp
SachinMChougale: Also move this code to module/function which keeps dfs related things in separate block

PR #723, opened 2026-06-30
RDKBWIFI-483: Add EM+ service type support

nefiguclu:
Controller detects EM+ agents via supported service TLV in autoconfig
  search and propagates the flag through dm_commit. Autoconfig response,
  topology response, autoconfig search, and BSS config request messages
  include EM+ service type when peer/self is EM+ capable. Agent reads
  EM_Plus field from /nvram/EasymeshPlus.json at startup to enable EM+
  mode; file absence means standard mode.

  DPP chirp path propagates peer_is_emplus through the full
  ec_manager → ec_configurator → send_autoconf_resp chain. Controller
  also persists is_emplus_agent via set_db_cfg_param on re-registration.

  Enable EM+ on agent: echo '{"EM_Plus": true}' > /nvram/EasymeshPlus.json

  0xA0 for EM+ controller
  0xA1 for EM+ agent

Issue: RDKBWIFI-483

### DIFF
--- a/src/agent/em_agent.cpp
+++ b/src/agent/em_agent.cpp
@@ -1514,6 +1514,8 @@ void em_agent_t::input_listener()
 
     em_printfout("bus open success");
 
+    load_em_plus_cfg();
+
     memset(&data, 0, sizeof(raw_data_t));
 
     while ((bus_error_val = desc->bus_data_get_fn(&m_bus_hdl, WIFI_WEBCONFIG_INIT_DML_DATA, &data)) != bus_error_success) {
@@ -2346,6 +2348,51 @@ bool em_agent_t::try_create_default_em_cfg(std::string interface)
     free(json_str);
 
     return true;
+}
+
+void em_agent_t::load_em_plus_cfg()
+{
+
+    m_data_model.get_device_info()->is_emplus_agent = 0;
+
+    FILE *fp = fopen(EM_PLUS_FILE, "r");
+    if (fp == NULL) {
+        return;
+    }
+
+    if (fseek(fp, 0, SEEK_END) != 0) {
+        fclose(fp);
+        return;
+    }
+    long sz = ftell(fp);
+    rewind(fp);
+
+    if (sz <= 0) {
+        fclose(fp);
+        return;
+    }
+
+    char *buf = static_cast<char *>(malloc(static_cast<size_t>(sz) + 1));
+    if (buf == NULL) {
+        fclose(fp);
+        return;
+    }
+    size_t nread = fread(buf, 1, static_cast<size_t>(sz), fp);
+    buf[nread] = '\0';
+    fclose(fp);
+
+    cJSON *root = cJSON_Parse(buf);
+    free(buf);
+    if (root == NULL) {
+      return;
+    }
+
+    cJSON *item = cJSON_GetObjectItem(root, "EM_Plus");
+    if (cJSON_IsBool(item) && cJSON_IsTrue(item)) {
+        m_data_model.get_device_info()->is_emplus_agent = 1;
+        em_printfout("EM+ agent mode enabled");
+    }
+    cJSON_Delete(root);
 }
 
 bool em_agent_t::try_start_dpp_onboarding()  {
--- a/src/em/config/em_configuration.cpp
+++ b/src/em/config/em_configuration.cpp
@@ -1456,14 +1456,34 @@ int em_configuration_t::send_topology_response_msg(unsigned char *dst, unsigned 
     len += static_cast<unsigned int> (sizeof (em_tlv_t) + tlv_len);
 
     // supported service tlv 17.2.1
-    tlv = reinterpret_cast<em_tlv_t *> (tmp);
-    tlv->type = em_tlv_type_supported_service;
-    tlv->len = htons(sizeof(em_enum_type_t) + 1);
-    tlv->value[0] = 1;
-    memcpy(&tlv->value[1], &service_type, sizeof(em_enum_type_t));
-
-    tmp += (sizeof(em_tlv_t) + sizeof(em_enum_type_t) + 1);
-    len += static_cast<unsigned int> (sizeof(em_tlv_t) + sizeof(em_enum_type_t) + 1);
+    {
+        uint8_t is_emplus = 0;
+        if (service_type == em_service_type_ctrl) {
+            dm_easy_mesh_t *agent_dm = get_mgr()->get_data_model(GLOBAL_NET_ID, hdr->src);
+            is_emplus = agent_dm && agent_dm->get_device_info()->is_emplus_agent;
+        } else {
+            is_emplus = dm->get_device_info()->is_emplus_agent;
+        }
+        tlv = reinterpret_cast<em_tlv_t *> (tmp);
+        tlv->type = em_tlv_type_supported_service;
+        if (is_emplus) {
+            unsigned char emplus_svc = (service_type == em_service_type_ctrl)
+                ? static_cast<unsigned char>(em_service_type_emplus_ctrl)
+                : static_cast<unsigned char>(em_service_type_emplus_agent);
+            tlv->len = htons(3);
+            tlv->value[0] = 2;
+            tlv->value[1] = static_cast<unsigned char>(service_type);
+            tlv->value[2] = emplus_svc;
+            tmp += (sizeof(em_tlv_t) + 3);
+            len += static_cast<unsigned int>(sizeof(em_tlv_t) + 3);
+        } else {
+            tlv->len = htons(sizeof(em_enum_type_t) + 1);
+            tlv->value[0] = 1;
+            memcpy(&tlv->value[1], &service_type, sizeof(em_enum_type_t));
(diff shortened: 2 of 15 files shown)


src/em/config/em_configuration.cpp
```diff
@@ -3264,10 +3284,14 @@ int em_configuration_t::create_bss_config_req_msg(uint8_t *buff, uint8_t dest_al
...
-    //  One SupportedService TLV.
-    // 1 service type followed by the service type value
-    uint8_t service_type_buff[2] = {1, service_type};
-    tmp = em_msg_t::add_tlv(tmp, &len, em_tlv_type_supported_service, service_type_buff, sizeof(service_type_buff));
+    //  One SupportedService TLV.
+    if (get_data_model()->get_device_info()->is_emplus_agent) {
+        uint8_t service_type_buff[3] = {2, static_cast<uint8_t>(service_type), static_cast<uint8_t>(em_service_type_emplus_agent)};
+        tmp = em_msg_t::add_tlv(tmp, &len, em_tlv_type_supported_service, service_type_buff, sizeof(service_type_buff));
+    } else {
+        uint8_t service_type_buff[2] = {1, static_cast<uint8_t>(service_type)};
```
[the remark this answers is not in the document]
nefiguclu: In the handle_bss_config_req_msg function, the service type information is currently being parsed but not used. Therefore, it is not necessary to obtain the EM+ information from there at this stage. There is also no risk in terms of byte length

src/em/config/em_configuration.cpp
```diff
@@ -5713,6 +5750,19 @@ int em_configuration_t::handle_autoconfig_search(unsigned char *buff, unsigned i
...
+        if (em_msg_t(buff + (sizeof(em_raw_hdr_t) + sizeof(em_cmdu_t)), len - static_cast<unsigned int>(sizeof(em_raw_hdr_t) + sizeof(em_cmdu_t))).get_supported_service(&svc)) {
+            for (unsigned int i = 0; i < svc.num && i < EM_MAX_SERVICE; i++) {
+                if (svc.service[i] == em_service_type_emplus_agent) {
+                    peer_is_emplus = true;
+                    break;
+                }
+            }
+        }
+    }
+
```
Nikita-Hakai, line 5875: can the previous update to `is_emplus_agent` in `find_em_for_msg_type()` be used here instead

src/em/em_msg.cpp
```diff
@@ -86,6 +86,40 @@ bool em_msg_t::get_al_mac_address(unsigned char *mac)
     return false;
 }
 
+bool em_msg_t::get_supported_service(em_supported_service_t *svc)
+{
+
```
Nikita-Hakai: extra line can be removed

src/em/config/em_configuration.cpp
nefiguclu: The is_emplus_agent flag obtained in the find_em_for_msg_type function is set in the data model through a parallel execution path. Another parallel flow reaches this point, so there is a potential race condition. For this reason, I preferred to extract it again here instead of relying on the value from the data model.

src/em/em_msg.cpp
```diff
@@ -86,6 +86,42 @@ bool em_msg_t::get_al_mac_address(unsigned char *mac)
...
+            svc->num = tlv->value[0];
+            if (svc->num > copy_len) {
+                svc->num = static_cast<unsigned char>(copy_len);
+            }
+            memset(svc->service, 0, EM_MAX_SERVICE);
+            memcpy(svc->service, &tlv->value[1], svc->num);
+            return true;
+        }
+        len -= static_cast<unsigned int>(sizeof(em_tlv_t) + tlv_len);
+        tlv = reinterpret_cast<em_tlv_t *>(reinterpret_cast<unsigned char *>(tlv) + sizeof(em_tlv_t) + tlv_len);
```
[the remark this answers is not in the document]
nefiguclu: This is already checked

src/agent/em_agent.cpp
```diff
@@ -2348,6 +2350,51 @@ bool em_agent_t::try_create_default_em_cfg(std::string interface)
...
+    }
+
+    char *buf = static_cast<char *>(malloc(static_cast<size_t>(sz) + 1));
+    if (buf == NULL) {
+        fclose(fp);
+        return;
+    }
+    size_t nread = fread(buf, 1, static_cast<size_t>(sz), fp);
+    buf[nread] = '\0';
+    fclose(fp);
```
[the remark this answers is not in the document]
nefiguclu: buf[nread] = '\0' ensures only the actually read bytes are null-terminated, so a partial fread results in cJSON_Parse failing gracefully and the function returning with is_emplus_agent = 0 the safe default. /nvram is a protected partition, so an oversized file isn't a realistic concern here. Leaving this out of scope for this PR.

src/agent/em_agent.cpp
```diff
@@ -2348,6 +2350,51 @@ bool em_agent_t::try_create_default_em_cfg(std::string interface)
...
+    cJSON *root = cJSON_Parse(buf);
+    free(buf);
+    if (root == NULL) {
+      return;
+    }
+
+    cJSON *item = cJSON_GetObjectItem(root, "EM_Plus");
+    if (cJSON_IsBool(item) && cJSON_IsTrue(item)) {
+        m_data_model.get_device_info()->is_emplus_agent = 1;
+        printf("%s:%d: EM+ agent mode enabled\n", __func__, __LINE__);
```
nikunjshah-cap: Use em_printfout instead

src/em/config/em_configuration.cpp
```diff
@@ -1456,14 +1457,34 @@ int em_configuration_t::send_topology_response_msg(unsigned char *dst, unsigned
...
-    tlv->type = em_tlv_type_supported_service;
-    tlv->len = htons(sizeof(em_enum_type_t) + 1);
-    tlv->value[0] = 1;
-    memcpy(&tlv->value[1], &service_type, sizeof(em_enum_type_t));
-
-    tmp += (sizeof(em_tlv_t) + sizeof(em_enum_type_t) + 1);
-    len += static_cast<unsigned int> (sizeof(em_tlv_t) + sizeof(em_enum_type_t) + 1);
+    {
+        uint8_t is_emplus = 0;
+        if (get_service_type() == em_service_type_ctrl) {
```
nikunjshah-cap: Can use the local variable service_type instead

src/em/config/em_configuration.cpp
```diff
@@ -1456,14 +1457,34 @@ int em_configuration_t::send_topology_response_msg(unsigned char *dst, unsigned
...
+            dm_easy_mesh_t *agent_dm = get_mgr()->get_data_model(GLOBAL_NET_ID, hdr->src);
+            is_emplus = agent_dm && agent_dm->get_device_info()->is_emplus_agent;
+        } else {
+            is_emplus = dm->get_device_info()->is_emplus_agent;
+        }
+        tlv = reinterpret_cast<em_tlv_t *> (tmp);
+        tlv->type = em_tlv_type_supported_service;
+        if (is_emplus) {
+            unsigned char emplus_svc = (get_service_type() == em_service_type_ctrl)
```
nikunjshah-cap: Can directly use local variable service_type instead

src/em/config/em_configuration.cpp
```diff
@@ -4023,12 +4047,20 @@ int em_configuration_t::create_autoconfig_resp_msg(unsigned char* buff, em_freq_
     // supported service tlv 17.2.1
     tlv = reinterpret_cast<em_tlv_t *> (tmp);
     tlv->type = em_tlv_type_supported_service;
-    tlv->len = htons(sizeof(em_enum_type_t) + 1);
-    tlv->value[0] = 1;
-    memcpy(&tlv->value[1], &service_type, sizeof(em_enum_type_t));
-
-    tmp += (sizeof(em_tlv_t) + sizeof(em_enum_type_t) + 1);
-    len += static_cast<int> (sizeof(em_tlv_t) + sizeof(em_enum_type_t) + 1);
+    if (peer_is_emplus) {
```
nikunjshah-cap, line 4086: Here, we explicitly check whether the peer is EM+ before reporting the EM+ service. However, similar checks do not appear to be present for other message types (e.g., Topology Response or BSS Configuration Request). Is there a specific reason for this difference in handling?

nefiguclu: Autoconfig Response is the only case where the Controller advertises the peer's EM+ status without being able to read it from a DM, the AL EM node handling Autoconfig Search is bound to the Controller's own DM (em_ctrl.cpp:770, em = al_em at :855), and the agent's DM may not even exist yet since dm_commit is queued asynchronously (em_mgr.cpp:119). Hence the explicit parameter.

Topology Response does have the equivalent check (em_configuration.cpp:1565-1571), it just looks the agent's DM up via hdr->src instead, since by then the DM exists.

BSS Config Request and Autoconfig Search are sent by the Agent advertising its own capability from its local DM, so there's no peer status to check.

Nikita-Hakai (approved):

nikunjshah-cap (approved):

PR #726, opened 2026-07-01
Nullptr checks on unified-wifi-mesh

ap934:


src/dm/dm_bsta_mld.cpp
```diff
@@ -97,8 +100,9 @@ bool dm_bsta_mld_t::operator == (const dm_bsta_mld_t& obj)
 
 dm_bsta_mld_t::dm_bsta_mld_t(em_bsta_mld_info_t *bsta_mld_info)
```
vlad-safonov, line 103: More idiomatic way is to use member initializer list, here and in other constructors, like this
`: m_bsta_mld_info{}`
(this is already done for `dm_bsta_mld_t::dm_bsta_mld_t()`),
so don't need to do memset 0

src/em/prov/easyconnect/ec_enrollee.cpp
```diff
@@ -582,6 +582,10 @@ bool ec_enrollee_t::handle_recfg_auth_confirm(ec_frame_t *frame, size_t len, uin
 
 bool ec_enrollee_t::handle_auth_request(ec_frame_t *frame, size_t len, uint8_t src_mac[ETHER_ADDR_LEN], unsigned int recv_freq)
 {
+    if (frame == nullptr || src_mac == nullptr) {
+        em_printfout("handle_auth_request: invalid nullptr argument");
```
vlad-safonov: `em_printfout` prints function and line. I can suggest to add `Error:` instead, so it will be easier to catch if something will be wrong

tests/main.cpp
```diff
@@ -21,698 +21,41 @@
...
-    "ec_pa_configurator_t.process_proxy_encap_dpp_msg_no_gas_session:"
-    "ec_pa_configurator_t.process_proxy_encap_dpp_msg_no_enrollee_mac:"
-    "ec_manager_t.handle_recv_ec_action_frame_invalid_frame:"
-    "ec_manager_t.handle_recv_ec_action_frame_invalid_MAC:"
-    "ec_manager_t.handle_recv_gas_pub_action_frame_null_MAC:"
-    "ec_manager_t.process_direct_encap_dpp_msg_enrollee_null_src_mac:"
-    "em_ctrl_t_Test.input_listen_default:"
-    "em_ctrl_t_Test.delete_nodes_valid:"
-    "em_ctrl_t_Test.delete_node_existing_ieee80211n_24";
+    // Exclude tests that cause segfaults due to deliberately passing null
```
vlad-safonov, line 24: If we want completely change logic of default filter may be we can also check if there is user provided filter from `argv` and use it instead of default one?

PR #733, opened 2026-07-06
RDKBWIFI-498: Fix memory leaks in EM agent and controller

dkyncu:
Both processes leak memory continuously during normal operation; worst observed case grew the agent from 28 MB to 399 MB in ~21 minutes and ended in an OOM reboot. Leaks were located with LeakSanitizer and validated fixed on a device (both processes RSS-flat over 8 hours).

Agent / command lifecycle:

- em_orch_t::handle_timeout: `ret` accumulated across all active commands, so one non-fini command blocked destroy_command() for every other command and the active queue grew without bound. Evaluate per command and drop the one-destroy-per-tick `break`.

- ~em_cmd_t never called deinit(), leaking the embedded data model of every command deleted outside destroy_command() (e.g. the per-radio ap_metrics_report commands, every reporting interval). deinit() is now idempotent and drains m_em_candidates before queue_destroy.

- dm_easy_mesh_t::deinit ignored the values returned by hash_map_remove, leaking every dm_sta_t/dm_scan_result_t on every deinit. Delete them; clone_hash_maps now deep-copies values so data models never share ownership.

- analyze_sta_list/sta_link_metrics/beacon_report/link_report leaked their stack-local data model (hash maps, m_wifi_data, decoded sta entries) on every bus event; call dm.deinit() before returning.

Controller:

- DB helpers (is_table_empty, compare_row, load_table, compare_db) never called close_result(); the leaked `select *` result set grew with the table on every sta metrics update cycle.

- tr_181_t::get_network_topology and em_ctrl_t:: publish_network_topology leaked the full topology cJSON tree (and the printed string on the publish path) on every call.

- tr_181_t::raw_data_set orphaned the bus_data_prop chain head it copies for the bus response; free it after the copy.


src/dm/dm_sta_list.cpp
```diff
@@ -352,10 +352,12 @@ bool dm_sta_list_t::compare_db(db_client_t& db_client, const dm_sta_t& sta)
         dm_easy_mesh_t::unhex(static_cast<unsigned int>(strlen(frame_body)), frame_body, EM_MAX_FRAME_BODY_LEN, info.frame_body);
 
         if (memcmp(static_cast<const void*>(&sta.m_sta_info), static_cast<const void*>(&info), sizeof(em_sta_info_t)) == 0) {
+            db_client.close_result(ctx);
             return true;
         }
```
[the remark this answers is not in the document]
Nikita-Hakai: same comment, couldn't find the definition to close_result().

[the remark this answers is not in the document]
dkyncu: my bad. close_result() comes from a vendor patch on the tree (mxl puma8) I first debugged this on. dropped those calls in the update. The rest of the PR is platform independent lifecycle fixes in common code, and the remaining DB change (compare_db() early-return drain) leaks upstream as well.

src/ctrl/tr_181/wfa_data_model/tr_181.cpp
```diff
@@ -901,8 +898,13 @@ bus_error_t tr_181_t::get_network_topology(char *event_name, raw_data_t *p_data,
...
-    em_printfout(" get_network_topology: node mac len: %d", sizeof(mac_addr_str_t));
+    cJSON_Delete(parent);
+    if (str == NULL) {
+        em_printfout("Memory allocation is failed");
+        return bus_error_out_of_resources;
+    }
 
+    p_data->data_type = bus_data_type_string;
     p_data->raw_data.bytes = reinterpret_cast<unsigned char *> (str);
     p_data->raw_data_len = static_cast<unsigned int> (strlen(str));
```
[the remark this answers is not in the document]
dkyncu: This preserves the exact behavior of the original code (the line was moved, not changed). the previous implementation also set raw_data_len = strlen(str) without the NUL. changing the payload length semantics is out of scope for this leak fix PR

Nikita-Hakai (dismissed):

src/dm/dm_easy_mesh.cpp
```diff
@@ -3008,7 +3008,9 @@ void dm_easy_mesh_t::clone_hash_maps(dm_easy_mesh_t& obj)
         macbytes_to_string(sta->m_sta_info.bssid, bss_mac_str);
         macbytes_to_string(sta->m_sta_info.radiomac, radio_mac_str);
         snprintf(key, sizeof(em_long_string_t), "%s@%s@%s", sta_mac_str, bss_mac_str, radio_mac_str);
-        hash_map_put(obj.m_sta_map, strdup(key),sta);
+        // deep copy: each data model's deinit() deletes its own map values,
+        // sharing the pointer with obj would double free
+        hash_map_put(obj.m_sta_map, strdup(key), new dm_sta_t(*sta));
```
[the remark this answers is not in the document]
dkyncu: added remove+delete before each insert in all three maps. Duplicates were already freed safely by the reworked deinit(), but deduplicating at insert is cleaner; clone_hash_maps is now idempotent.

src/dm/dm_easy_mesh.cpp
```diff
@@ -3028,7 +3030,7 @@ void dm_easy_mesh_t::clone_hash_maps(dm_easy_mesh_t& obj)
         macbytes_to_string(sta->m_sta_info.bssid, bss_mac_str);
         macbytes_to_string(sta->m_sta_info.radiomac, radio_mac_str);
         snprintf(key, sizeof(em_long_string_t), "%s@%s@%s", sta_mac_str, bss_mac_str, radio_mac_str);
-        hash_map_put(obj.m_sta_dassoc_map, strdup(key),sta);
+        hash_map_put(obj.m_sta_dassoc_map, strdup(key), new dm_sta_t(*sta));
         sta = static_cast<dm_sta_t *> (hash_map_get_next(m_sta_dassoc_map, sta));
```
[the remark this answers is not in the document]
dkyncu: same

src/dm/dm_easy_mesh.cpp
```diff
@@ -3018,7 +3020,7 @@ void dm_easy_mesh_t::clone_hash_maps(dm_easy_mesh_t& obj)
         macbytes_to_string(sta->m_sta_info.bssid, bss_mac_str);
         macbytes_to_string(sta->m_sta_info.radiomac, radio_mac_str);
         snprintf(key, sizeof(em_long_string_t), "%s@%s@%s", sta_mac_str, bss_mac_str, radio_mac_str);
-        hash_map_put(obj.m_sta_assoc_map, strdup(key),sta);
+        hash_map_put(obj.m_sta_assoc_map, strdup(key), new dm_sta_t(*sta));
         sta = static_cast<dm_sta_t *> (hash_map_get_next(m_sta_assoc_map, sta));
```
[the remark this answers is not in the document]
dkyncu: same

vlad-safonov:
One test is failed with this PR
```
❌ CloneHashMapsTest_positive_CloneMaps
	/__w/unified-wifi-mesh/unified-wifi-mesh/easymesh_project/unified-wifi-mesh/build/ctrl/../..//tests/test_l1_dm_easy_mesh.cpp:21791
	Expected equality of these values:
	  dst_sta
	    Which is: 0x522000043900
	  src_sta
	    Which is: 0x522000040900
```

src/ctrl/tr_181/wfa_data_model/tr_181.cpp
```diff
@@ -928,8 +925,13 @@ bus_error_t tr_181_t::get_network_topology(char *event_name, raw_data_t *p_data,
     dm_ctrl->get_network_config(parent, const_cast<char*>(GLOBAL_NET_ID));
 
     str = cJSON_Print(parent);
-    em_printfout(" get_network_topology: node mac len: %d", sizeof(mac_addr_str_t));
+    cJSON_Delete(parent);
+    if (str == NULL) {
+        em_printfout("Memory allocation is failed");
```
vlad-safonov:
```suggestion
        em_printfout("Error: Memory allocation is failed");
```

src/ctrl/tr_181/wfa_data_model/tr_181.cpp
```diff
@@ -928,8 +925,13 @@ bus_error_t tr_181_t::get_network_topology(char *event_name, raw_data_t *p_data,
...
+        return bus_error_out_of_resources;
+    }
 
+    p_data->data_type = bus_data_type_string;
     p_data->raw_data.bytes = reinterpret_cast<unsigned char *> (str);
     p_data->raw_data_len = static_cast<unsigned int> (strlen(str));
```
vlad-safonov: Actually we have fixed a lot of places in OneWifi by adding this +1 to `p_data->raw_data_len` as part of others topics, if this is only one place probably it is worth to fix it here as well

src/dm/dm_easy_mesh.cpp
```diff
@@ -3042,23 +3047,31 @@ void dm_easy_mesh_t::deinit()
     em_2xlong_string_t key;
     mac_addr_str_t dev_mac_str, radio_mac_str, bss_mac_str, sta_mac_str, scanner_mac_str;
 
+    // deinit must be idempotent: ~em_cmd_t calls it and some paths call it
+    // explicitly before delete, so guard each map and reset it to NULL
+
     //destroy elements of m_scan_result_map
+    if (m_scan_result_map != NULL) {
 	res = static_cast<dm_scan_result_t *> (hash_map_get_first(m_scan_result_map));
```
vlad-safonov: as this is very important function may be we can fix indentations here and below for it to make in cleaner (somebody added if's but didn't move internal loops).
like 

```suggestion
    if (m_scan_result_map != NULL) {
	    res = static_cast<dm_scan_result_t *> (hash_map_get_first(m_scan_result_map));
	    ...
```
It will help in reading and understanding this code

src/dm/dm_easy_mesh.cpp
```diff
@@ -3008,7 +3008,10 @@ void dm_easy_mesh_t::clone_hash_maps(dm_easy_mesh_t& obj)
         macbytes_to_string(sta->m_sta_info.bssid, bss_mac_str);
         macbytes_to_string(sta->m_sta_info.radiomac, radio_mac_str);
         snprintf(key, sizeof(em_long_string_t), "%s@%s@%s", sta_mac_str, bss_mac_str, radio_mac_str);
-        hash_map_put(obj.m_sta_map, strdup(key),sta);
+        // deep copy: each data model's deinit() deletes its own map values,
```
nikunjshah-cap: In the existing implementation, both the original and cloned hashmaps reference the same underlying objects. As a result, updates made to an object's members through one hashmap are visible through the other as well. With this change, that behavior will no longer hold, as the cloned hashmap will contain independent object instances.

Have we assessed the impact of this behavioral change on existing flows? It would be good to verify that no logic currently relies on shared object references between the original and cloned hashmaps. 

We should also review and update the relevant test cases to reflect the new semantics. Although the updates can be handled in a separate pull request if preferred.

src/dm/dm_easy_mesh.cpp
```diff
@@ -3042,23 +3047,31 @@ void dm_easy_mesh_t::deinit()
...
 					tmp_res->m_scan_result.id.op_class, tmp_res->m_scan_result.id.channel, tmp_res->m_scan_result.id.scanner_type);
-		hash_map_remove(m_scan_result_map, key);
+		// hash_map_remove frees only the map node and returns the value
+		delete static_cast<dm_scan_result_t *> (hash_map_remove(m_scan_result_map, key));
 	}
 
-	hash_map_destroy(m_scan_result_map);	
+	hash_map_destroy(m_scan_result_map);
+	m_scan_result_map = NULL;
+    }
```
nikunjshah-cap, line 3100: For consistency, assign NULL to local variable sta: `sta = NULL;`

nikunjshah-cap (changes requested):

src/dm/dm_easy_mesh.cpp
vlad-safonov: Hi @nikunjshah-cap , as a related improvement - what do you think about using `std::unordered_map` instead of OneWifi hash_map?

nikunjshah-cap: Since these maps contain the list of associated/disassociated stations and their metrics, using std::unordered_map does seem reasonable from a container perspective. 

However, we should explicitly define and preserve the intended copy/clone behavior while migrating. Test coverage should also be reviewed accordingly.

vlad-safonov:
Question: if enable em controller ASAN like this
`EXTRA_OECONF:append = " EM_ASAN_CTRL=true"` and after boot check `em_ctrl.service` logs 
`journalctl -u em_ctrl`
is it clean?
We have ASAN reports
```
1970 Jan 01 00:00:50 rdkb-lgm sh[3941]: Starting /usr/ccsp/EasyMesh/onewifi_em_ctrl_pre_start.sh
1970 Jan 01 00:00:51 rdkb-lgm systemd[1]: Started EasyMesh Controller service.
1970 Jan 01 00:01:49 rdkb-lgm sh[4178]: =================================================================
1970 Jan 01 00:01:49 rdkb-lgm sh[4178]: ==4178==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61c00003277b at pc 0x7f95556fb0f5 bp 0x7f954d6b1ef0 sp 0x7f954d6b16a8
....
```
pointing to `em_capability_t::handle_ap_cap_report(unsigned char*, unsigned int)`
(we use older version of EM ~17-June so this already can be fixed)

Nikita-Hakai:
Hello @dkyncu, can you please check on the latest comments

src/ctrl/tr_181/wfa_data_model/tr_181.cpp
dkyncu: done

src/ctrl/tr_181/wfa_data_model/tr_181.cpp
dkyncu: applied. raw_data_len now includes the NUL, consistent with raw_data_set(const char *)

src/dm/dm_easy_mesh.cpp
dkyncu: done

src/dm/dm_easy_mesh.cpp
dkyncu: std::unordered_map with explicit copy semantics would prevent this bug class, and the migration needs its own copy behavior definition and test pass. wider refactor though, so best tracked as a separate work item

src/dm/dm_easy_mesh.cpp
dkyncu: done

dkyncu:
> Question: if enable em controller ASAN like this `EXTRA_OECONF:append = " EM_ASAN_CTRL=true"` and after boot check `em_ctrl.service` logs `journalctl -u em_ctrl` is it clean? We have ASAN reports
> 
> ```
> 1970 Jan 01 00:00:50 rdkb-lgm sh[3941]: Starting /usr/ccsp/EasyMesh/onewifi_em_ctrl_pre_start.sh
> 1970 Jan 01 00:00:51 rdkb-lgm systemd[1]: Started EasyMesh Controller service.
> 1970 Jan 01 00:01:49 rdkb-lgm sh[4178]: =================================================================
> 1970 Jan 01 00:01:49 rdkb-lgm sh[4178]: ==4178==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61c00003277b at pc 0x7f95556fb0f5 bp 0x7f954d6b1ef0 sp 0x7f954d6b16a8
> ....
> ```
> 
> pointing to `em_capability_t::handle_ap_cap_report(unsigned char*, unsigned int)` (we use older version of EM ~17-June so this already can be fixed)

This PR is scoped to the leaks we actually caught (LSan on both processes, 8h RSS flat on device); ASAN's overflow checks weren't part of this work. The overflow you hit is a pre existing issue in the AP Capability Report TLV parsing, unrelated to these changes, and several parsers in that path changed after your June tree (#749/#750), so it is worth retesting on current develop.

Sanitizer findings are inherently tied to the code paths exercised. different message flows would surface different issues. so rather than trying to cover everything in one PR, the practical approach is to fix findings incrementally as they are caught, each in its own change, as done here.

Nikita-Hakai (approved):

vlad-safonov:
> > Question: if enable em controller ASAN like this `EXTRA_OECONF:append = " EM_ASAN_CTRL=true"` and after boot check `em_ctrl.service` logs `journalctl -u em_ctrl` is it clean? We have ASAN reports
> > ```
> > 1970 Jan 01 00:00:50 rdkb-lgm sh[3941]: Starting /usr/ccsp/EasyMesh/onewifi_em_ctrl_pre_start.sh
> > 1970 Jan 01 00:00:51 rdkb-lgm systemd[1]: Started EasyMesh Controller service.
> > 1970 Jan 01 00:01:49 rdkb-lgm sh[4178]: =================================================================
> > 1970 Jan 01 00:01:49 rdkb-lgm sh[4178]: ==4178==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61c00003277b at pc 0x7f95556fb0f5 bp 0x7f954d6b1ef0 sp 0x7f954d6b16a8
> > ....
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > pointing to `em_capability_t::handle_ap_cap_report(unsigned char*, unsigned int)` (we use older version of EM ~17-June so this already can be fixed)
> 
> This PR is scoped to the leaks we actually caught (LSan on both processes, 8h RSS flat on device); ASAN's overflow checks weren't part of this work. The overflow you hit is a pre existing issue in the AP Capability Report TLV parsing, unrelated to these changes, and several parsers in that path changed after your June tree (#749/#750), so it is worth retesting on current develop.
> 
> Sanitizer findings are inherently tied to the code paths exercised. different message flows would surface different issues. so rather than trying to cover everything in one PR, the practical approach is to fix findings incrementally as they are caught, each in its own change, as done here.

Yes agreed, originally I was thinking it is introduced by this PR

vlad-safonov (approved):

src/dm/dm_easy_mesh.cpp
nikunjshah-cap: Agreed that the wider refactoring can be discussed and addressed separately.

For this PR, could you share what verification was done to assess the impact of the changed clone behavior? Since the original and cloned hashmaps will no longer share object references, it would be good to confirm that no existing functional flows rely on updates being reflected across both maps.

If any impact is identified, we can evaluate whether it should be addressed as part of this PR or tracked separately under a different issue.

dkyncu: @nikunjshah-cap 

Verified statically: clone_hash_maps() has exactly two callers (dm_easy_mesh_agent.cpp - analyze_sta_list() :222, analyze_sta_link_metrics() :785). In both, the clone source is a stack local working model destroyed via dm.deinit() right after the clone, in the same function. so no flow can rely on updates being reflected across the maps, the source doesn't outlive the call. The command's maps are then only read (report TLV building).

Note also that dm_easy_mesh_t::operator= already deep copies m_sta_map (new dm_sta_t per entry). the shared references in clone_hash_maps() were the inconsistency; this change aligns it with the copy semantics the class already uses everywhere else. With sharing kept, dm.deinit() would in fact double free against the command's own teardown.

Runtime: 8h+ on device RSS flat with assoc/disassoc churn and periodic link metric reporting exercising these maps every interval; the updated unit test encodes the new semantics.

nikunjshah-cap (approved):

dkyncu:
> ## Pull request overview
> Copilot reviewed 9 out of 9 changed files in this pull request and generated no new comments.
> 
> Suppressed comments (1)
> **tests/test_l1_dm_easy_mesh.cpp:21909**
> 
> * The deep-copy test currently validates equality using only the STA MAC (`m_sta_info.id`). Since `clone_hash_maps()` is intended to deep-copy the full `dm_sta_t` contents, this can miss regressions where other `em_sta_info_t` fields are not copied correctly. Consider comparing the full `em_sta_info_t` (or using `dm_sta_t::operator==`) for both maps.

The copy constructor memcpy's the whole em_sta_info_t, so partial field copying isn't a failure mode here; the test asserts the property this change introduces, a distinct object per map. A content comparison would only add value if those copy semantics change.

PR #753, opened 2026-07-18
RDKBWIFI-521: Remove extra closing brace from DataElements schema

dkyncu:
Data_Elements_JSON_Schema_v3.0.json has an extra closing brace in the
Radio object, shifting everything after it (NetworkSSIDList,
MSCS/SCSDisallowedStaList, notification) out of their "properties"
objects. Registration walks "properties" only, so
Device.WiFi.DataElements.Network.SSID.{i} was never registered.
Remove the extra brace.

### DIFF
--- a/src/ctrl/tr_181/wfa_data_model/Data_Elements_JSON_Schema_v3.0.json
+++ b/src/ctrl/tr_181/wfa_data_model/Data_Elements_JSON_Schema_v3.0.json
@@ -3065,7 +3065,6 @@
                                             "description": "List of 802.11 standards in operation for this radio instance",
                                             "enum": ["a", "b", "g", "n", "ac", "ax", "be", "bn"]
                                         }
-                                        }
                                     }
                                 }
                             }


src/ctrl/tr_181/wfa_data_model/Data_Elements_JSON_Schema_v3.0.json
```diff
@@ -3070,6 +3070,12 @@
                                 }
                             }
                         }
+                    },
```
Nikita-Hakai: Hello @dkyncu , I don't think we can modify the schema file as we use the same copy of Schema json available from the specification of "Wi-Fi_Data_Elements_Specification_Package_v3.0". If the network ssid list is not getting registered on the bus, then we might have to look into the code which sends for the registration by parsing the schema? possibly a bug?

dkyncu: Hi @Nikita-Hakai, thanks for the comment. dug into this further. The root cause is an extra closing brace in the schema's Radio object, which shifts NetworkSSIDList (and everything after it) out of "properties", so registration never saw Network.SSID. Updated the PR to just remove that extra brace.

Nikita-Hakai (approved):

Nikita-Hakai (approved):

PR #758, opened 2026-07-22
RDKBWIFI-527: SetSSID: implement the AKMsAllowed input argument

dkyncu:
The BBF TR-181 Device.WiFi.DataElements.Network.SetSSID() command defines an AKMsAllowed input argument (since 2.17, akm_t enumeration) to select the AKM suites / security mode of an SSID, but the rbus method implementation only parses SSID, AddRemoveChange, PassPhrase, Band and HaulType; AKMsAllowed is silently ignored and the method still returns Success.

Implement the AKMsAllowed input for Add and Change: validate the value, update the target entry's AKMsAllowed and derive its AuthType ("WPA2 Personal" / "WPA3 Personal" / "WPA3 Transition"), then let the existing SetSSID flow (decode -> autoconfig renew -> M2) apply it.

Only the akm_t values that map to a WSC M2 auth type are accepted (psk, sae, psk+sae). The dpp* and SuiteSelector enumeration values are rejected as invalid input for now, since there is no plumbing from a NetworkSSID entry to a DPP configuration object / suite selector; support can be added when that path exists.

### DIFF
--- a/inc/tr_181.h
+++ b/inc/tr_181.h
@@ -89,6 +89,7 @@ static const yang_to_tr181_map g_yang_map[] = {
 #define TR181_BAND_MAX_LEN         16
 #define TR181_ADDREMOVE_MAX_LEN    16
 #define TR181_HAULTYPE_MAX_LEN     32
+#define TR181_AKMS_MAX_LEN         32
 #define TR181_CHLIST_MAX_LEN       128
 #define TR181_BSSID_MAX_LEN        32
 #define TR181_REQMODE_MAX_LEN      24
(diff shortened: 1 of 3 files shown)


nikunjshah-cap (approved):
This change maps the AKMsAllowed parameter received through SetSSID() to the corresponding authentication type. A few considerations:

- Although TR‑181 defines AKMsAllowed as an array, the current implementation supports only a single value.
- As described, WPA2, WPA3, and WPA2/WPA3 Transition modes are only supported

Nikita-Hakai (approved):

PR #762, opened 2026-07-27
RDKBWIFI-530: EasyMesh-one_wifi_em_ctrl high cpu usage and Device.WiFi.DataElements timeout

vineeth-shankar28:
Root Cause:
The EasyMesh controller was incorrectly included in the DataElements Network.Device list together with agents. This created gaps in the DM IDs, causing the device enumeration logic to repeatedly traverse the DM list while searching for sequential IDs. As a result, one_wifi_em_ctrl consumed high CPU and Device.WiFi.DataElements DM queries took several minutes or timed out.

Fix:
- Skip controller entries during DataElements device enumeration.
- Track the maximum valid DM ID and limit the iteration range.
- Ignore controller entries while building the Network.Device table.

This prevents unnecessary repeated traversals, reduces CPU utilization, and allows Device.WiFi.DataElements queries to complete successfully.


vineeth-shankar28:
Hi @Nikita-Hakai 
Why am I not seeing any copilot reviews here?

Nikita-Hakai (approved):

PR #767, opened 2026-08-03
RDKBWIFI-542: Populate STA ClientCapabilities and WiFi6Capabilities

dkyncu:
STA.{i}.ClientCapabilities is defined by TR-181 as the base64 encoded frame body of the last (Re)Association Request, but the getters emitted an empty string even though the frame body is available in the STA info. The STA WiFi6Capabilities.{param} leaves were registered from the schema with no getter binding, so they always read false and 802.11ax stations were never reported as HE-capable.

Emit ClientCapabilities as base64(frame body) and bind the WiFi6Capabilities leaves, deriving their values from the HE Capabilities element (ext id 35) in the stored frame body; members with no HE counterpart keep reading false.


src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -5600,6 +5600,88 @@ char* dm_easy_mesh_ctrl_t::get_sta_vht_caps_str(char *vht_cap_hex, char *buf, si
...
+/* Locate the HE Capabilities element (ext id 35) in the sta's stored (re)assoc frame body. */
+bool dm_easy_mesh_ctrl_t::find_sta_he_caps(em_sta_info_t *si, const unsigned char **mac_caps,
+    const unsigned char **phy_caps)
+{
+    unsigned int offset = 0;
+
+    if ((si == NULL) || (si->frame_body_len == 0)) {
+        return false;
+    }
+    while (offset + 2 <= si->frame_body_len) {
```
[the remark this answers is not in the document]
dkyncu: Aligned this with the existing frame body handling: find_sta_he_caps() now skips the fixed fields (capab_info + listen_interval) the same way dm_sta_t::decode_sta_capability() does. Full offset handling that also covers reassociation bodies (6 extra fixed bytes, a limitation shared by decode_sta_capability() today) is coming in an upcoming patchset that makes IE location robust for all (re)assoc frame layouts.

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -5600,6 +5600,88 @@ char* dm_easy_mesh_ctrl_t::get_sta_vht_caps_str(char *vht_cap_hex, char *buf, si
...
+        unsigned char id = si->frame_body[offset];
+        unsigned char len = si->frame_body[offset + 1];
+        if (offset + 2 + len > si->frame_body_len) {
+            break;
+        }
+        /* ext id(1) + HE MAC caps(6) + HE PHY caps(11) + HE-MCS/NSS(>=4) */
+        if ((id == 255) && (len >= 18) && (si->frame_body[offset + 2] == 35)) {
```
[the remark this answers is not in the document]
dkyncu: Tightened as suggested: the HE Capabilities element must now be at least 22 bytes (ext id + MAC caps + PHY caps + the mandatory HE-MCS/NSS set), so a truncated element is no longer treated as valid.

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -5600,6 +5600,89 @@ char* dm_easy_mesh_ctrl_t::get_sta_vht_caps_str(char *vht_cap_hex, char *buf, si
...
+        if (offset + 2 + len > si->frame_body_len) {
+            break;
+        }
+        /* ext id(1) + HE MAC caps(6) + HE PHY caps(11) + HE-MCS/NSS(>=4) */
+        if ((id == 255) && (len >= 22) && (si->frame_body[offset + 2] == 35)) {
+            *mac_caps = &si->frame_body[offset + 3];
+            *phy_caps = &si->frame_body[offset + 9];
+            return true;
+        }
+        offset += 2 + static_cast<unsigned int>(len);
```
Nikita-Hakai: can we maybe use named constants or macros for the magic numbers

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -5600,6 +5600,89 @@ char* dm_easy_mesh_ctrl_t::get_sta_vht_caps_str(char *vht_cap_hex, char *buf, si
...
+};
+
+/* Derive WiFi6Capabilities.<param> from the HE caps element; false if param is not a member. */
+bool dm_easy_mesh_ctrl_t::sta_wifi6_cap_value(em_sta_info_t *si, const char *param, bool *out)
+{
+    const unsigned char *m = NULL, *p = NULL;
+    unsigned int i;
+    bool member = false;
+
+    for (i = 0; i < sizeof(sta_wifi6_cap_members) / sizeof(sta_wifi6_cap_members[0]); i++) {
```
Nikita-Hakai, line 5817: null checks for params, out etc

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -5600,6 +5600,89 @@ char* dm_easy_mesh_ctrl_t::get_sta_vht_caps_str(char *vht_cap_hex, char *buf, si
...
+    "TWTResponder", "SpatialReuse", "AnticipatedChannelUsage",
+};
+
+/* Derive WiFi6Capabilities.<param> from the HE caps element; false if param is not a member. */
+bool dm_easy_mesh_ctrl_t::sta_wifi6_cap_value(em_sta_info_t *si, const char *param, bool *out)
+{
+    const unsigned char *m = NULL, *p = NULL;
+    unsigned int i;
+    bool member = false;
+
```
Nikita-Hakai, line 5812:
```suggestion

    if (!si || !param || !out) {
        return false;
    }
```

src/ctrl/dm_easy_mesh_ctrl.cpp
```diff
@@ -5600,6 +5600,89 @@ char* dm_easy_mesh_ctrl_t::get_sta_vht_caps_str(char *vht_cap_hex, char *buf, si
...
+    } else if (strcmp(param, "Beamformee80orLess") == 0) {
+        *out = ((p[4] >> 2) & 0x07) != 0; /* beamformee STS <= 80 MHz */
+    } else if (strcmp(param, "BeamformeeAbove80") == 0) {
+        *out = ((p[4] >> 5) & 0x07) != 0; /* beamformee STS > 80 MHz */
+    } else if (strcmp(param, "ULMUMIMO") == 0) {
+        *out = (p[2] & 0xc0) != 0; /* full/partial bandwidth UL MU-MIMO */
+    } else if (strcmp(param, "TWTRequestor") == 0) {
+        *out = (m[0] & 0x02) != 0; /* MAC B1 */
+    } else if (strcmp(param, "TWTResponder") == 0) {
+        *out = (m[0] & 0x04) != 0; /* MAC B2 */
```
Nikita-Hakai: constants/macros can be used here too I believe

src/ctrl/dm_easy_mesh_ctrl.cpp
dkyncu: Done

src/ctrl/dm_easy_mesh_ctrl.cpp
dkyncu: Done

src/ctrl/dm_easy_mesh_ctrl.cpp
dkyncu: Done

src/ctrl/dm_easy_mesh_ctrl.cpp
dkyncu: Done

Nikita-Hakai (approved):

PR #770, opened 2026-08-06
RDKBWIFI-554: Fix collocated EasyMesh onboarding radios remaining on default SSID

Navya-Vemula:
RDKBWIFI-554: Fix collocated EasyMesh onboarding radios remaining on default SSID

Issue:
Collocated EasyMesh onboarding intermittently fails, causing one or more radios to remain on the default SSID.

Root Cause:
AL-SAP uses SOCK_STREAM sockets, which do not preserve message boundaries. The receiver was assuming that a single recv() call would return the complete packet, causing incorrect packet processing during onboarding.

Fix:
Implemented AL-SAP stream socket packet length based framing to read the packet length header first and receive the complete payload before processing.

TEST:
Verified successful collocated EasyMesh onboarding with agent and client connections.
Confirmed radios receive the provisioned SSID and maintain connectivity after onboarding.
 
 Signed-off-by: Navya Vemula <navya.navya@capgemini.com>

### DIFF
--- a/src/al-sap/al_service_access_point.cpp
+++ b/src/al-sap/al_service_access_point.cpp
@@ -1,6 +1,29 @@
 #include "al_service_access_point.h"
 #include "al_service_utils.h"
 
+namespace {
+// Reads exactly len bytes from fd into buffer starting at offset, retrying on partial reads and EINTR.
+void recvExact(int fd, unsigned char *buffer, size_t len) {
+    size_t totalRead = 0;
+    while (totalRead < len) {
+        ssize_t bytesRead = recv(fd, buffer + totalRead, len - totalRead, 0);
+        if (bytesRead == 0) {
+            throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
+        }
+        if (bytesRead < 0) {
+            if (errno == EINTR) {
+                continue;
+            }
+            if (errno == EBADF || errno == ECONNRESET) {
+                throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
+            }
+            throw AlServiceException("Failed to receive message through Unix socket", PrimitiveError::IndicationFailed);
+        }
+        totalRead += static_cast<size_t>(bytesRead);
+    }
+}
+}
+
 // Constructor: Connects to the Unix domain socket using the provided path --> moved from hardcoded to check in the unit test for socket creation
 AlServiceAccessPoint::AlServiceAccessPoint(const std::string &dataSocketPath, const std::string &controlSocketPath) : alDataSocketpath(dataSocketPath),
                                                                                                                       alControlSocketpath(controlSocketPath)
@@ -202,20 +225,19 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
     int fragmentId = 0;
     bool receivingFragments = true;
     AlServiceDataUnit message;
-
+    constexpr size_t lengthPrefixSize = sizeof(uint32_t);
     while (receivingFragments) {
-        std::vector<unsigned char> buffer(SOCKET_MTU, 0x00);
+        std::vector<unsigned char> buffer(lengthPrefixSize, 0x00);
+        buffer.reserve(SOCKET_MTU);
 
         // Receive data from the socket
-        ssize_t bytesRead = recv(alDataSocketDescriptor, buffer.data(), buffer.size(), 0);
-        if (bytesRead <= 0) {
-            if (errno == EBADF || errno == ECONNRESET) {
-                throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
-            }
-            throw AlServiceException("Failed to receive message through Unix socket", PrimitiveError::IndicationFailed);
+        recvExact(alDataSocketDescriptor, buffer.data(), lengthPrefixSize);
+        uint32_t packet_size = convert_bytes_into_u32(buffer);
+        if (packet_size < (PACKET_HEADER_SIZE - lengthPrefixSize) || packet_size > (SOCKET_MTU - lengthPrefixSize)) {
+             throw AlServiceException("Invalid packet size", PrimitiveError::InvalidMessage);
         }
-        buffer.resize(bytesRead);
-
+        buffer.resize(lengthPrefixSize + packet_size);
+        recvExact(alDataSocketDescriptor, buffer.data() + lengthPrefixSize, packet_size);
         // Deserialize the received fragment
         AlServiceDataUnit fragment;
         try {
@@ -225,7 +247,7 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
         }
         #ifdef DEBUG_MODE
         std::cout << "Received fragment " << static_cast<int>(fragment.getFragmentId())
-                  << " - Size: " << bytesRead << " bytes, "
+                  << " - Size: " << buffer.size() << " bytes, "
                   << "isFragment: " << static_cast<int>(fragment.getIsFragment()) << ", "
                   << "FragmentId: " << static_cast<int>(fragment.getFragmentId()) << ", "
                   << "isLastFragment: " << static_cast<int>(fragment.getIsLastFragment()) << std::endl;
@@ -233,7 +255,7 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
         // Check if this is a single message (non-fragmented)
         if (fragment.getIsFragment() == 0 && fragment.getIsLastFragment() == 1) {
             #ifdef DEBUG_MODE
-            std::cout << "Received a non-fragmented message of size: " << bytesRead << " bytes." << std::endl;
+            std::cout << "Received a non-fragmented message of size: " << buffer.size() << " bytes." << std::endl;
             #endif
             return fragment; // Return immediately, as no reassembly is needed
         }
--- a/src/al-sap/al_service_utils.cpp
+++ b/src/al-sap/al_service_utils.cpp
@@ -77,7 +77,8 @@ uint32_t convert_bytes_into_u32(const std::vector<unsigned char>& bytes)
 {
     if (bytes.size() < 4)
         return 0;
-    return static_cast<uint32_t>((bytes[0] << 24) | (bytes[1] << 16) | (bytes[2] << 8) | (bytes[3] << 0));
+    return (static_cast<uint32_t>(bytes[0]) << 24) | (static_cast<uint32_t>(bytes[1]) << 16) |
+           (static_cast<uint32_t>(bytes[2]) << 8)  | static_cast<uint32_t>(bytes[3]);
 }
 
 std::vector<unsigned char> remove_length_delimited_part(const std::vector<unsigned char>& buffer)


Navya-Vemula:

src/al-sap/al_service_access_point.cpp
```diff
@@ -204,17 +204,34 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
...
 
+        buffer.reserve(SOCKET_MTU);
+        buffer.resize(4);
         // Receive data from the socket
-        ssize_t bytesRead = recv(alDataSocketDescriptor, buffer.data(), buffer.size(), 0);
+        bytesRead = recv(alDataSocketDescriptor, buffer.data(), 4 , 0);
+        if (bytesRead <= 0) {
+            if (errno == EBADF || errno == ECONNRESET) {
+                throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
+            }
```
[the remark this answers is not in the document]
Navya-Vemula: addressed review comment

src/al-sap/al_service_access_point.cpp
```diff
@@ -204,17 +204,47 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
...
-            if (errno == EBADF || errno == ECONNRESET) {
-                throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
+        do {
+            bytesRead = recv(alDataSocketDescriptor, buffer.data() + totalRead, 4 - totalRead, 0);
+            if (bytesRead <= 0) {
+                 if (errno == EBADF || errno == ECONNRESET) {
+                     throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
+                 }
+                 throw AlServiceException("Failed to receive message through Unix socket", PrimitiveError::IndicationFailed);
             }
```
DharmalakshmiA: Here, we first read the frame length and then read the corresponding packet data based on that length.

However, if two packets are already available in the socket buffer:

`len + data` → First packet
`len + data` → Second packet

the current logic reads only the first packet. The second packet remains unread in the socket buffer and could be missed/dropped depending on how the socket handling is implemented.

Navya-Vemula: I had actually observed this scenario in the previous implementation. From the tcpdump capture, I could see 3 separate M2 messages (~2290 bytes each). However, in the old implementation, a single recv() was sometimes returning ~4590 bytes, which indicates that two M2 messages were coalesced into a single stream read.

This is the reason for introducing the framing logic. We first read the 4-byte length header and then read exactly the corresponding payload size. So if the socket buffer 
contains:

**len1 + data1 + len2 + data2**

the current iteration consumes only:

**len1 + data1**

and the remaining:

**len2 + data2**

stays in the socket receive buffer. It is then processed in the next iteration when serviceAccessPointDataIndication() is called again.

I verified this with the new logs where each M2 is received independently:

- Packet Size received from header = 2291
- Payload accumulated = 2291/2291
-  
- Packet Size received from header = 2291
- Payload accumulated = 2291/2291
-  
- Packet Size received from header = 2291
- Payload accumulated = 2291/2291

So the second packet is not dropped. It remains in the socket buffer and is read and processed in the subsequent iteration. This fix was specifically added to handle the packet coalescing behavior that I observed in the previous implementation.

vlad-safonov:
Problems with the new framing code
1. Debug logging is now wrong (real bug introduced by this diff).
bytesRead is reused as the loop variable in both do...while loops. After the loops finish, it holds only the return value of the last partial recv() call, not the packet/fragment total. The old code logged the true single-recv() size; now the #ifdef DEBUG_MODE logs ("Received fragment ... Size: " << bytesRead, "Received a non-fragmented message of size: " << bytesRead) print a meaningless number.

2. Duplicated recv loop (violates DRY, doubles the bug surface).
The exact same "read N bytes, retry on partial read, throw on error" logic is written out twice (once for the 4-byte length header, once for the payload). Any future fix (e.g. EINTR handling) has to be made in two places and is easy to miss in one.

3. recv() == 0 (peer graceful close) is misclassified.
bytesRead <= 0 is treated as an error and only mapped to SocketClosed if errno happens to be EBADF/ECONNRESET. On an orderly shutdown, recv() returns 0 and doesn't set errno, so a normal disconnect during a partial read throws the generic IndicationFailed instead of SocketClosed. This bug pre-existed for the single old recv(), but is now duplicated.

4. No EINTR retry. A signal interrupting recv() returns -1/EINTR, which this code treats as a hard failure instead of retrying.

5. Manual big-endian decode duplicates an existing utility. al_service_utils.h already declares convert_bytes_into_u32() (used nowhere else consistently), but the new code hand-rolls the same 4-byte shift/OR logic inline instead of reusing it.

6. Magic numbers. The literal 4 (length-prefix size) appears 6+ times; PACKET_HEADER_SIZE - 4 is a fragile way to express "header size without the length field" — if PACKET_HEADER_SIZE's definition ever changes, this silently changes meaning.

7. No recovery/resync on "Invalid packet size". Once a bad length is read, the exception is thrown but the socket isn't drained or closed — any bytes the peer already sent for that message remain queued, permanently desyncing the framing for the connection's remaining lifetime.

8. Unbounded blocking read (DoS/hang risk). There's no SO_RCVTIMEO/poll-with-timeout — if a peer sends a valid length header but then stalls (slow-loris), serviceAccessPointDataIndication() blocks forever with no way to time out.

9. No test coverage. There's no test_l1_al_service_access_point.cpp; this new framing/looping logic isn't exercised by any unit test, unlike sibling classes (al_service_data_unit, al_service_utils, etc.).

vlad-safonov:
we can try something like this, but need to review carefully
```
diff --git a/src/al-sap/al_service_access_point.cpp b/src/al-sap/al_service_access_point.cpp
index 2743b50..0ee101e 100644
--- a/src/al-sap/al_service_access_point.cpp
+++ b/src/al-sap/al_service_access_point.cpp
@@ -1,6 +1,29 @@
 #include "al_service_access_point.h"
 #include "al_service_utils.h"

+namespace {
+// Reads exactly len bytes from fd into buffer starting at offset, retrying on partial reads and EINTR.
+void recvExact(int fd, unsigned char *buffer, size_t len) {
+    size_t totalRead = 0;
+    while (totalRead < len) {
+        ssize_t bytesRead = recv(fd, buffer + totalRead, len - totalRead, 0);
+        if (bytesRead == 0) {
+            throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
+        }
+        if (bytesRead < 0) {
+            if (errno == EINTR) {
+                continue;
+            }
+            if (errno == EBADF || errno == ECONNRESET) {
+                throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
+            }
+            throw AlServiceException("Failed to receive message through Unix socket", PrimitiveError::IndicationFailed);
+        }
+        totalRead += static_cast<size_t>(bytesRead);
+    }
+}
+}
+
 // Constructor: Connects to the Unix domain socket using the provided path --> moved from hardcoded to check in the unit test for socket creation
 AlServiceAccessPoint::AlServiceAccessPoint(const std::string &dataSocketPath, const std::string &controlSocketPath) : alDataSocketpath(dataSocketPath),
                                                                                                                       alControlSocketpath(controlSocketPath)
@@ -203,48 +226,21 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
     bool receivingFragments = true;
     AlServiceDataUnit message;

-    while (receivingFragments) {
-        std::vector<unsigned char> buffer;
-        ssize_t bytesRead;
-        uint32_t packet_size;
-        size_t totalRead = 0;
+    constexpr size_t lengthPrefixSize = sizeof(uint32_t);

+    while (receivingFragments) {
+        // Read the length prefix, then the payload it describes (SOCK_STREAM doesn't preserve message boundaries).
+        std::vector<unsigned char> buffer(lengthPrefixSize, 0x00);
         buffer.reserve(SOCKET_MTU);
-        buffer.resize(4);
-        // Receive data from the socket
-        do {
-            bytesRead = recv(alDataSocketDescriptor, buffer.data() + totalRead, 4 - totalRead, 0);
-            if (bytesRead <= 0) {
-                 if (errno == EBADF || errno == ECONNRESET) {
-                     throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
-                 }
-                 throw AlServiceException("Failed to receive message through Unix socket", PrimitiveError::IndicationFailed);
-            }
-            totalRead += bytesRead;
-        } while (totalRead < 4);
-
-        packet_size =
-            (static_cast<uint32_t>(buffer[0]) << 24) |
-            (static_cast<uint32_t>(buffer[1]) << 16) |
-            (static_cast<uint32_t>(buffer[2]) << 8)  |
-            (static_cast<uint32_t>(buffer[3]));
+        recvExact(alDataSocketDescriptor, buffer.data(), lengthPrefixSize);

-        if (packet_size < (PACKET_HEADER_SIZE - 4) || packet_size > (SOCKET_MTU - 4)) {
-            throw AlServiceException( "Invalid packet size", PrimitiveError::InvalidMessage);
+        uint32_t packet_size = convert_bytes_into_u32(buffer);
+        if (packet_size < (PACKET_HEADER_SIZE - lengthPrefixSize) || packet_size > (SOCKET_MTU - lengthPrefixSize)) {
+            throw AlServiceException("Invalid packet size", PrimitiveError::InvalidMessage);
         }

-        buffer.resize(packet_size + 4);
-        totalRead = 0;
-        do {
-            bytesRead = recv(alDataSocketDescriptor, buffer.data() + 4 + totalRead, packet_size - totalRead, 0);
-            if (bytesRead <= 0) {
-                if (errno == EBADF || errno == ECONNRESET) {
-                    throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
-                }
-                throw AlServiceException("Failed to receive message through Unix socket", PrimitiveError::IndicationFailed);
-            }
-            totalRead += bytesRead;
-        } while (totalRead < packet_size);
+        buffer.resize(lengthPrefixSize + packet_size);
+        recvExact(alDataSocketDescriptor, buffer.data() + lengthPrefixSize, packet_size);

         // Deserialize the received fragment
         AlServiceDataUnit fragment;
@@ -255,7 +251,7 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
         }
         #ifdef DEBUG_MODE
         std::cout << "Received fragment " << static_cast<int>(fragment.getFragmentId())
-                  << " - Size: " << bytesRead << " bytes, "
+                  << " - Size: " << buffer.size() << " bytes, "
                   << "isFragment: " << static_cast<int>(fragment.getIsFragment()) << ", "
                   << "FragmentId: " << static_cast<int>(fragment.getFragmentId()) << ", "
                   << "isLastFragment: " << static_cast<int>(fragment.getIsLastFragment()) << std::endl;
@@ -263,7 +259,7 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
         // Check if this is a single message (non-fragmented)
         if (fragment.getIsFragment() == 0 && fragment.getIsLastFragment() == 1) {
             #ifdef DEBUG_MODE
-            std::cout << "Received a non-fragmented message of size: " << bytesRead << " bytes." << std::endl;
+            std::cout << "Received a non-fragmented message of size: " << buffer.size() << " bytes." << std::endl;
             #endif
             return fragment; // Return immediately, as no reassembly is needed
         }
diff --git a/src/al-sap/al_service_utils.cpp b/src/al-sap/al_service_utils.cpp
index c35bdc6..bd1ff4a 100644
--- a/src/al-sap/al_service_utils.cpp
+++ b/src/al-sap/al_service_utils.cpp
@@ -77,7 +77,8 @@ uint32_t convert_bytes_into_u32(const std::vector<unsigned char>& bytes)
 {
     if (bytes.size() < 4)
         return 0;
-    return static_cast<uint32_t>((bytes[0] << 24) | (bytes[1] << 16) | (bytes[2] << 8) | (bytes[3] << 0));
+    return (static_cast<uint32_t>(bytes[0]) << 24) | (static_cast<uint32_t>(bytes[1]) << 16) |
+           (static_cast<uint32_t>(bytes[2]) << 8)  | static_cast<uint32_t>(bytes[3]);
 }

 std::vector<unsigned char> remove_length_delimited_part(const std::vector<unsigned char>& buffer)
```

Navya-Vemula:
hi @vlad-safonov,

Thanks for the detailed review. I've addressed the review comments by introducing a common recvExact() helper to remove duplicated receive logic, adding handling for recv() == 0 and EINTR, replacing the inline length decoding with convert_bytes_into_u32(), fixing the debug logging size reporting, and replacing hardcoded length-prefix values with lengthPrefixSize.

Nikita-Hakai:
Awaiting test results from @DharmalakshmiA

src/al-sap/al_service_access_point.cpp
```diff
@@ -202,20 +225,19 @@ AlServiceDataUnit AlServiceAccessPoint::serviceAccessPointDataIndication() {
...
+        buffer.reserve(SOCKET_MTU);
 
         // Receive data from the socket
-        ssize_t bytesRead = recv(alDataSocketDescriptor, buffer.data(), buffer.size(), 0);
-        if (bytesRead <= 0) {
-            if (errno == EBADF || errno == ECONNRESET) {
-                throw AlServiceException("Socket closed or connection reset", PrimitiveError::SocketClosed);
-            }
-            throw AlServiceException("Failed to receive message through Unix socket", PrimitiveError::IndicationFailed);
+        recvExact(alDataSocketDescriptor, buffer.data(), lengthPrefixSize);
```
DharmalakshmiA, line 234: we are receiving the length of the packet and trying to read the frame based on the length.
What if 2 frames were stored in the socket buffer ? 
Ex: M1_Radio2_len | M1_Radio2_frame | M1_Radio3_len | M1_Radio3_data

Navya-Vemula: recvExact() reads exactly the number of bytes requested and returns once that amount has been received.

For example, if the socket buffer contains:

**M1_Radio2_len | M1_Radio2_frame | M1_Radio3_len | M1_Radio3_frame**


the first call:

recvExact(fd, ..., lengthPrefixSize);

consumes only M1_Radio2_len.

After decoding the length, the second call:

recvExact(fd, ..., packet_size);

consumes only M1_Radio2_frame.

At that point, M1_Radio3_len | M1_Radio3_frame remain in the socket receive buffer and are processed during the next iteration of the loop.

Since this is a SOCK_STREAM socket, message boundaries are reconstructed using the length prefix. Multiple frames being present in the socket buffer is expected and is handled correctly by the framing logic.

DharmalakshmiA: At that point, M1_Radio3_len | M1_Radio3_frame remain in the socket receive buffer and are processed during the next iteration of the loop.

What happens when another packet arrives in the socket ? Again the socket buffer contains 
M1_Radio3_len | M1_Radio3_frame | Next_packet_len | Next_packet_frame.
In this case, we may miss few frames.

Navya-Vemula: I verified this scenario using runtime logs.

Each invocation reads exactly one length header and one payload.

Example logs:

> recvExact: requested=4 received=4
> Packet size from header = 53
> recvExact: requested=53 received=53
> Payload fully received: 53 bytes


After the frame is processed (proto_process returned), the next invocation starts and immediately reads another valid length header.

Example logs :

> proto_process returned
> Waiting on serviceAccessPointDataIndication
> Reading length header
> recvExact: requested=4 received=4
> Packet size from header = 53

 
This pattern repeats continuously for consecutive packets of varying sizes (53, 2316, 44, 93, 1492, 1432, etc.), and each packet is reconstructed and passed to proto_process() successfully.

If M1_Radio3_frame were skipped when Next_packet arrived, the subsequent read would not start at the correct frame boundary. We would expect invalid packet lengths, deserialization failures, or reconstruction errors. None of these were observed in the logs.

Based on the runtime validation, I was unable to reproduce a scenario where queued frames are missed when additional packets are appended to the socket buffer.

DharmalakshmiA: We previously encountered a scenario during testing where two M1 packets were appended in the socket buffer.

However, we are currently unable to reproduce the issue, even in our setup.

Could you please check whether incorporating these changes would help handle the packet-append scenario, in case it occurs again?
@Nikita-Hakai  FYI

https://github.com/rdkcentral/unified-wifi-mesh/commit/48632c641d2108a770cf1a196044d21380db6802
https://github.com/rdkcentral/unified-wifi-mesh/commit/b5c1d885209ad2d94d0ecc424a9d26eefd9a8aa1

Navya-Vemula: Hi @DharmalakshmiA,

Could you please let me know if you're still seeing any concerns with the current implementation and share the test results/observations from your validation?

In parallel, I am also reviewing the packet-append scenario against the commits you referenced (48632c6 and b5c1d88).

PR #771, opened 2026-08-10
RDKBWIFI-555: Locate IEs in (re)association request frame bodies

dkyncu:
decode_sta_capability() and find_sta_he_caps() reach the IEs of the STA's stored (Re)Association Request by skipping a fixed 4 octets, the capability information and listen interval fields. That holds for an association request only.

A reassociation request carries the Current AP address as well, so its IEs start 6 octets further in. For a reassociating STA the walk therefore begins inside that address and every element after it is read out of alignment, which leaves the parsed capabilities and the WiFi6Capabilities HE lookup working on misparsed elements. A producer that stores the IEs without the fixed fields shifts the walk the other way.

Probe the association, reassociation and IEs only layouts instead, preferring the offset that opens with the mandatory SSID and Supported Rates elements and whose element walk consumes the body exactly, so that a Current AP address beginning with 0x00 cannot pass for an SSID element, and use that offset in both parsers.

Cover the association, reassociation with a Current AP address that starts with 0x00, IEs only, truncated and unrecognised layouts with unit tests.

### DIFF
--- a/src/dm/dm_sta.cpp
+++ b/src/dm/dm_sta.cpp
@@ -392,6 +392,60 @@ void dm_sta_t::parse_sta_bss_radio_from_key(const char *key, mac_address_t sta, 
 
 }
 
+/* A (Re)Association Request opens with the mandatory SSID element followed by
+ * the mandatory Supported Rates element, so a candidate offset has to hold that
+ * pair. Requiring the pair rather than the SSID element alone keeps a Current AP
+ * address such as 00:04:.. from passing as an SSID element. */
+static bool starts_with_ssid_and_rates(const unsigned char *body, unsigned int len, unsigned int off)
+{
+    unsigned int next;
+
+    if ((off + EM_IE_HDR_LEN > len) || (body[off] != EM_EID_SSID) ||
+        (body[off + 1] > (EM_MAX_SSID_LEN - 1))) {
+        return false;
+    }
+    next = off + EM_IE_HDR_LEN + static_cast<unsigned int>(body[off + 1]);
+
+    return ((next + EM_IE_HDR_LEN <= len) && (body[next] == EM_EID_SUPP_RATES));
+}
+
+/* The frame body may carry the 802.11 fixed fields before the IEs, or the IEs
+ * alone. The first pass takes the offset that opens with those two elements and
+ * whose element walk consumes the body exactly. A stored body may be truncated,
+ * in which case no walk ends on the length, so the second pass keeps the same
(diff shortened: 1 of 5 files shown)


src/dm/dm_sta.cpp
```diff
@@ -392,6 +392,43 @@ void dm_sta_t::parse_sta_bss_radio_from_key(const char *key, mac_address_t sta,
...
+        }
+    }
+    for (i = 0; i < sizeof(offsets) / sizeof(offsets[0]); i++) {
+        off = offsets[i];
+        if ((off + EM_IE_HDR_LEN <= len) && (body[off] == EM_EID_SSID) &&
+            (body[off + 1] <= (EM_MAX_SSID_LEN - 1)) &&
+            (off + EM_IE_HDR_LEN + body[off + 1] <= len)) {
+            return off;
+        }
+    }
```
Nikita-Hakai, line 445: This loop (pass 2) doesn't match what the comment says. Comment says on truncation just assume offset 0 right? but the code still checks assoc/reassoc offsets first and returns those if they look kinda valid. Either make the fallback only check offset 0, or update the comment accordingly

src/dm/dm_sta.cpp
```diff
@@ -392,6 +392,43 @@ void dm_sta_t::parse_sta_bss_radio_from_key(const char *key, mac_address_t sta,
...
+    return 0;
```
Nikita-Hakai, line 446: it could mean "IEs actually start at byte 0" or "parsing failed and nothing matched." Is this not an issue in your use case?

src/dm/dm_sta.cpp
dkyncu: You're right, the comment was misleading. the code does what I intended. Falling back to offset 0 on a truncated body would be wrong: the correct offset is still the assoc/reassoc one, and parsing from 0 would treat the fixed fields as elements. the second pass keeps the same candidates and only drops the exact consumption requirement. I have reworded the comment to say exactly that.

src/dm/dm_sta.cpp
dkyncu: not an issue here. Offset 0 is a legitimate layout in its own right. Some agents store only the IEs, so it cannot be reserved as an error value. All callers bounds check every element read, so an unrecognised body yields no elements rather than an over read. An explicit "unknown" return would not change anything either: every caller would still skip the parse.

Nikita-Hakai (approved):

PR #775, opened 2026-08-16
RDKBWIFI-548: em_ctrl Core Dump Observed After Controller Boot-Up

chandraprakashduraisamy:
Added code changes to address RDKBWIFI-548: em_ctrl Core Dump Observed After Controller Boot-Up.

### DIFF
--- a/src/ctrl/em_ctrl.cpp
+++ b/src/ctrl/em_ctrl.cpp
@@ -1460,22 +1460,45 @@ em_ctrl_t::~em_ctrl_t()
 #ifdef AL_SAP
 AlServiceAccessPoint* em_ctrl_t::al_sap_register(const std::string& data_socket_path, const std::string& control_socket_path)
 {
-    AlServiceAccessPoint* sap = new AlServiceAccessPoint(data_socket_path.c_str(), control_socket_path.c_str());
+    AlServiceAccessPoint* sap = NULL;
+    try {
+        sap = new AlServiceAccessPoint(data_socket_path.c_str(), control_socket_path.c_str());
+        if (NULL == sap) {
+            em_printfout("%s-%d: Failed to allocate AlServiceAccessPoint", __func__, __LINE__);
+            return NULL;
+        }
 
-    AlServiceRegistrationRequest registrationRequest(SAPActivation::SAP_ENABLE, ServiceType::EmController);
-    sap->serviceAccessPointRegistrationRequest(registrationRequest);
+        AlServiceRegistrationRequest registrationRequest(SAPActivation::SAP_ENABLE, ServiceType::EmController);
+        sap->serviceAccessPointRegistrationRequest(registrationRequest);
 
-    AlServiceRegistrationResponse registrationResponse = sap->serviceAccessPointRegistrationResponse();
+        AlServiceRegistrationResponse registrationResponse = sap->serviceAccessPointRegistrationResponse();
 
-    RegistrationResult result = registrationResponse.getResult();
-    if (result == RegistrationResult::SUCCESS) {
-        g_al_mac_sap = registrationResponse.getAlMacAddressLocal();
-        uint8_t* al_mac_bytes = g_al_mac_sap.data();
-        em_printfout("AL SAP registration successful, AL MAC: %s", util::mac_to_string(al_mac_bytes).c_str());
-
-        m_data_model.set_dev_interface_mac(al_mac_bytes);
-    } else {
-        std::cout << "Registration failed with error: " << static_cast<int>(result) << std::endl;
+        RegistrationResult result = registrationResponse.getResult();
+        if (result == RegistrationResult::SUCCESS) {
+            g_al_mac_sap = registrationResponse.getAlMacAddressLocal();
+            uint8_t* al_mac_bytes = g_al_mac_sap.data();
+            if (NULL == al_mac_bytes) {
+                em_printfout("%s-%d: AL SAP registration failed with invalid AL MAC: %s", util::mac_to_string(al_mac_bytes).c_str());
+                delete sap;
+                return NULL;
+            }
+            em_printfout("AL SAP registration successful, AL MAC: %s", util::mac_to_string(al_mac_bytes).c_str());
+            m_data_model.set_dev_interface_mac(al_mac_bytes);
+        } else {
+            std::cout << "Registration failed with error: " << static_cast<int>(result) << std::endl;
+            delete sap;
+            return NULL;
+        }
+    }
+    catch (const AlServiceException& e) {
+        em_printfout("%s-%d: AL SAP registration exception %s", __func__, __LINE__, e.what());
+        delete sap;
+        return NULL;
+    }
+    catch (const std::exception e) {
+        em_printfout("%s-%d: Unknown exception %s during AL SAP registration", __func__, __LINE__, e.what());
+        delete sap;
+        return NULL;
     }
 
     return sap;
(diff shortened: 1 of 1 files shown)


src/ctrl/em_ctrl.cpp
```diff
@@ -1460,22 +1460,45 @@ em_ctrl_t::~em_ctrl_t()
...
 AlServiceAccessPoint* em_ctrl_t::al_sap_register(const std::string& data_socket_path, const std::string& control_socket_path)
 {
-    AlServiceAccessPoint* sap = new AlServiceAccessPoint(data_socket_path.c_str(), control_socket_path.c_str());
-
-    AlServiceRegistrationRequest registrationRequest(SAPActivation::SAP_ENABLE, ServiceType::EmController);
-    sap->serviceAccessPointRegistrationRequest(registrationRequest);
+    AlServiceAccessPoint* sap = NULL;
+    try {
+        sap = new AlServiceAccessPoint(data_socket_path.c_str(), control_socket_path.c_str());
+	if (NULL == sap) {
```
Nikita-Hakai: indentation is mismatched

src/ctrl/em_ctrl.cpp
```diff
@@ -1460,22 +1460,45 @@ em_ctrl_t::~em_ctrl_t()
...
+        AlServiceRegistrationResponse registrationResponse = sap->serviceAccessPointRegistrationResponse();
 
-        m_data_model.set_dev_interface_mac(al_mac_bytes);
-    } else {
-        std::cout << "Registration failed with error: " << static_cast<int>(result) << std::endl;
+        RegistrationResult result = registrationResponse.getResult();
+        if (result == RegistrationResult::SUCCESS) {
+            g_al_mac_sap = registrationResponse.getAlMacAddressLocal();
+            uint8_t* al_mac_bytes = g_al_mac_sap.data();
+            if (NULL == al_mac_bytes) {
```
vlad-safonov, line 1480: at which scenario `if (NULL == al_mac_bytes)` can be `true` keeping in mind that underlaying type is `std::array`?

src/ctrl/em_ctrl.cpp
```diff
@@ -1460,22 +1460,45 @@ em_ctrl_t::~em_ctrl_t()
...
+                em_printfout("%s-%d: AL SAP registration failed with invalid AL MAC: %s", util::mac_to_string(al_mac_bytes).c_str());
```
vlad-safonov, line 1481: and if assume that it will be NULL somehow what is intention to print something pointing to `NULL` + missed leading args for `%s-%d`? 
Seems like there is some mismatch with all that block

src/ctrl/em_ctrl.cpp
```diff
@@ -1460,22 +1460,45 @@ em_ctrl_t::~em_ctrl_t()
...
+	    delete sap;
+	    return NULL;
+        }
+    }
+    catch (const AlServiceException& e) {
+        em_printfout("%s-%d: AL SAP registration exception %s", __func__, __LINE__, e.what());
+        delete sap;
+        return NULL;
+    }
+    catch (const std::exception e) {
```
vlad-safonov, line 1498: copies by value unnecessarily, probably should be `const std::exception&`

src/ctrl/em_ctrl.cpp
```diff
@@ -1460,22 +1460,45 @@ em_ctrl_t::~em_ctrl_t()
...
+        em_printfout("%s-%d: Unknown exception %s during AL SAP registration", __func__, __LINE__, e.what());
```
vlad-safonov, line 1499: `em_printfout ` has by default ` __func__, __LINE__`

vlad-safonov, line 1499:
```suggestion
        em_printfout("Unknown exception %s during AL SAP registration", e.what());
```

src/ctrl/em_ctrl.cpp
```diff
@@ -1460,22 +1460,45 @@ em_ctrl_t::~em_ctrl_t()
 #ifdef AL_SAP
 AlServiceAccessPoint* em_ctrl_t::al_sap_register(const std::string& data_socket_path, const std::string& control_socket_path)
 {
-    AlServiceAccessPoint* sap = new AlServiceAccessPoint(data_socket_path.c_str(), control_socket_path.c_str());
-
-    AlServiceRegistrationRequest registrationRequest(SAPActivation::SAP_ENABLE, ServiceType::EmController);
-    sap->serviceAccessPointRegistrationRequest(registrationRequest);
+    AlServiceAccessPoint* sap = NULL;
+    try {
+        sap = new AlServiceAccessPoint(data_socket_path.c_str(), control_socket_path.c_str());
```
vlad-safonov, line 1465: operator `new` never returns null in C++. It throws `std::bad_alloc` on failure, which is already caught below. The check `if (NULL == sap)`  is dead code and misleading.

src/ctrl/em_ctrl.cpp
```diff
@@ -1460,22 +1460,45 @@ em_ctrl_t::~em_ctrl_t()
 #ifdef AL_SAP
 AlServiceAccessPoint* em_ctrl_t::al_sap_register(const std::string& data_socket_path, const std::string& control_socket_path)
 {
-    AlServiceAccessPoint* sap = new AlServiceAccessPoint(data_socket_path.c_str(), control_socket_path.c_str());
-
-    AlServiceRegistrationRequest registrationRequest(SAPActivation::SAP_ENABLE, ServiceType::EmController);
-    sap->serviceAccessPointRegistrationRequest(registrationRequest);
+    AlServiceAccessPoint* sap = NULL;
```
vlad-safonov, line 1463:
```suggestion
    AlServiceAccessPoint* sap = nullptr;
```

src/ctrl/em_ctrl.cpp
```diff
@@ -1488,7 +1511,23 @@ int main(int argc, const char *argv[])
 {
     em_ctrl_t  *em_ctrl = em_ctrl_t::get_em_ctrl_instance();
 #ifdef AL_SAP
-    g_sap = em_ctrl->al_sap_register("/tmp/al_em_ctrl_data_socket", "/tmp/al_em_ctrl_control_socket");
+    const char* data_socket_path = "/tmp/al_em_ctrl_data_socket";
+    const char* control_socket_path = "/tmp/al_em_ctrl_control_socket";
+
+    if(0 == access(data_socket_path, F_OK) && 0 == access(control_socket_path, F_OK)) {
```
vlad-safonov, line 1517: Checking for socket existence with `access()` before connecting is a Time-Of-Check To Time-Of-Use race - the sockets can disappear between the check and the `AlServiceAccessPoint` constructor call. 
The original code was fine: just call `al_sap_register` and let the constructor throw, which the catch handles. The pre-flight check adds no safety and blocks startup if sockets haven't been created yet (e.g., the service starts late).

src/ctrl/em_ctrl.cpp
```diff
@@ -1488,7 +1511,23 @@ int main(int argc, const char *argv[])
...
+        g_sap = em_ctrl->al_sap_register(data_socket_path, control_socket_path);
+        if (NULL == g_sap) {
+            em_printfout("%s-%d: Error in AL SAP registration, exiting", __func__, __LINE__);
```
vlad-safonov, line 1520: like explained above
```suggestion
            em_printfout("Error in AL SAP registration, exiting");
```

src/ctrl/em_ctrl.cpp
```diff
@@ -1488,7 +1511,23 @@ int main(int argc, const char *argv[])
...
+            return -1;
+        }
+    }
+    else {
+        em_printfout("%s-%d: Data Socket: %s, Control Socket: %s", __func__, __FILE__,
```
vlad-safonov, line 1525:
```suggestion
        em_printfout("Data Socket: %s, Control Socket: %s",
```

src/ctrl/em_ctrl.cpp
```diff
@@ -1488,7 +1511,23 @@ int main(int argc, const char *argv[])
...
+	              access(data_socket_path, F_OK) == 0 ? "present" : "missing",
+                      access(control_socket_path, F_OK) == 0 ? "present" : "missing");
+        em_printfout("%s-%d: Required AL SAP socket(s) not available, exiting", __func__, __LINE__);
```
vlad-safonov, line 1528:
```suggestion
        em_printfout("Required AL SAP socket(s) not available, exiting");
```

src/ctrl/em_ctrl.cpp
chandraprakashduraisamy: @Nikita-Hakai  Fixed the indentation and pushed the changes. Please review and share your comments/approval.
Thanks!

Nikita-Hakai (approved):

PR #764, opened 2026-07-29
RDKBWIFI-442:Beacon Report Query/Response.

prashant-singh1:
Beacon metrics query/response support and refactor MLO clients handling
Core Changes:
- Implement Beacon metrics query and response handling

Client Assoc/Disassoc Updates:
- For MLO clients, requent topo query to get latest new mld links info on a re-assoc as the links may differ
- If non-MLO client, do not re-request client cap. Also ensure to not delete caps on a disassoc.
- handle_assoc_sta_mld_topology_update() to reset m_sta_map entries before marking active links, replacing stale scan logic.
- Ensure DB disassoc sync: Write per-link associated=false entries to m_sta_assoc_map on disassociation events.
- Force re-assoc topo sync: Always trigger topo_sync for MLO STAs in analyze_sta_assoc_event() even if sta_exists=true.

Capability Updates:
- Consolidate per-link cap writes into dm_easy_mesh_t::apply_sta_cap_to_maps() to update m_sta_assoc_map and
  m_sta_map together for immediate datamodel update.
- Support 6GHz HE/EHT tags: Add decoding for extended tags (HE) and (EHT) in decode_sta_capability() to properly capture 6GHz capabilities.

PR #785, opened 2026-09-01
RDKBWIFI-585: Query capabilities again when a non-MLD client reconnects

dkyncu:
A non-MLD client that disassociates and associates again with the same BSS is treated as already known: the Topology Notification disassoc path keeps the stored association frame and the assoc path skips the Client Capability Query when a row with that frame exists. Nothing then marks the row associated again, so DataElements keeps hiding the station, and the capabilities derived from the frame (standards, SGI, spatial streams, bandwidth) stay those of the first association.

Only treat the client as known while its row is still associated. A reconnect after a disassoc carries a new (Re)Association frame, so it goes through the Client Capability Query like a new client; a repeated assoc notification for an active client is still not re-queried and the stored capabilities are still kept across a disassoc.

### DIFF
--- a/src/ctrl/dm_easy_mesh_ctrl.cpp
+++ b/src/ctrl/dm_easy_mesh_ctrl.cpp
@@ -2756,9 +2756,10 @@ int dm_easy_mesh_ctrl_t::analyze_sta_assoc_event(em_bus_event_t *evt, em_cmd_t *
                 // MLD STA is "known" only if we already have its capability data
                 sta_exists = (iter->m_sta_info.frame_body_len > 0);
             } else {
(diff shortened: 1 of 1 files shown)


rakhilpe:
@Nikita-Hakai  Change looks good to me.

dkyncu:
Hi @Nikita-Hakai @rakhilpe 

Rebased on develop. #764 already handles the disassoc side, so the change is now a single line in analyze_sta_assoc_event: a client whose row is no longer associated is queried for capabilities again on reconnect.

Please take another look.

Nikita-Hakai (approved):

PR #797, opened 2026-09-09
RDKBWIFI-592: Airties device info TLV for boot id and cloud credentials

meycint:
Reason for change: The controller has no way to learn an agent's boot id or
  its cloud communication credentials, as no WFA TLV carries them. An Airties
  vendor specific TLV (0x0003) is added to the AutoConfig WSC M1, carrying
  boot id, client id, client secret, product class and device role. The
  controller stores them on the device and exposes the first three over
  TR-181. The boot id is read from /tmp/emex_boot_id, or drawn from
  /dev/urandom and persisted there on the first boot; the credentials are read
  from Device.Services.X_AIRTIES_Edge.AuthConfig.ClientID and .ClientPassword.
  Product class and device role are hardcoded to gateway/controller until
  their sources are implemented.
  Two data model defects in the way are also fixed: dm_device_t::operator=
  copied 32 bytes of secondary device type over a 4 byte enum, and
  commit_config() dropped the new fields on every save.
Test Procedure: (RDKB) rbuscli may be used to read the new parameters:
  rbuscli get Device.WiFi.DataElements.Network.Device.1.X_AIRTIES_BootID
  rbuscli get Device.WiFi.DataElements.Network.Device.1.X_AIRTIES_ClientID
  rbuscli get Device.WiFi.DataElements.Network.Device.1.X_AIRTIES_ClientSecret
  They are only populated once an M1 carrying the TLV has been received, so
  they stay empty for an agent that does not send it.
Risks: Low. The TLV is optional and vendor specific, so an agent or
  controller without it interoperates unchanged.


Nikita-Hakai:
Hi @meycint , since there are growing vendor specific implementations do you think we can do something like this going forward, not necessarily in this PR:

```
src/em/vendor/
    em_vendor_rdk.h        em_vendor_rdk.cpp        (OUI d8:9c:8e — RDK's own extensions)
    em_vendor_airties.h    em_vendor_airties.cpp     (OUI 88:41:fc — Airties' extensions)
```

Each vendor file owns: its OUI constant, its private sub-type enum, every TLV builder/parser it currently has scattered across em_metrics.cpp, em_configuration.cpp, em_capability.cpp, em_policy_cfg.cpp, and one guarded entry point that does the OUI check exactly once.

which eventually is like this:

```
int em_configuration_t::handle_vendor_specific_tlv(unsigned char *buff, unsigned int len) {
    .....
    if (len < EM_VENDOR_OUI_SIZE + sizeof(uint16_t)) return -1;
    auto *vendor = reinterpret_cast<em_vendor_specific_v_t *>(buff);
    em_vendor_airties::handle_device_info_tlv(vendor->data, len - EM_VENDOR_OUI_SIZE, get_data_model());
    return 0;
}
```

I also see vendor specific parameters add-ons to existing members like device_info, radio_info etc. 

This is just one initial thought from my end and doesn't necessarily need to be followed—other ideas or architectures are totally welcome, and we can decide on the best path when we take this up. Please let me know what you think!

Thank you

PR #800, opened 2026-09-12
RDKCOM-5653: RDKBDEV-3505 Fix socket descriptor leaks in em_cmd_agent_t::execute()

Yashu37:
**Issue:**
In em_cmd_agent_t::execute(), socket file descriptors are used for listening and receiving client connections. When listen() or recv() fails, the function exits the corresponding failure path without properly closing the socket descriptor.
This can leave socket file descriptors open and result in a resource/file descriptor leak when these failures occur repeatedly.

**Steps to Reproduce:**
    1.Invoke em_cmd_agent_t::execute().
    2.Allow the listener socket to be created successfully.
    3.Trigger a listen() failure and observe that the listener socket is not closed before returning.
    4.Trigger a recv() failure after a client connection is accepted.
    5.Observe that m_dsock is not closed before exiting the receive loop.
    6.Repeated failures can cause socket file descriptors to remain open.

**Expected Behavior:**
When listen() or recv() fails, the corresponding socket file descriptor should be properly closed before the function exits the failure path.
Normal successful execution and existing logging/return behavior should remain unchanged.

**Fix:**
    1.Close lsock when listen() fails before returning from em_cmd_agent_t::execute().
    2.Use the existing close_listener_socket() cleanup helper for the listener failure path.
    3.Close m_dsock when recv() fails before breaking out of the receive loop.
    4.Preserve existing logging, return values, and successful socket-processing behavior.

**Acceptance Criteria:**
1.lsock is closed on the listen() failure path.
2.m_dsock is closed on the recv() failure path.
3.No socket file descriptor remains open on these failure paths.
4.Existing error logging is preserved.
5.Existing return and control-flow behavior is preserved.
6.Successful listen() and recv() behavior remains unchanged.
7.Relevant build/tests complete successfully with no regressions.


src/agent/em_cmd_agent.cpp
```diff
@@ -66,7 +66,8 @@ int em_cmd_agent_t::execute(em_long_string_t result)
 
     if ((ret = listen(lsock, 20)) == -1) {
         printf("%s:%d: listen error on socket, err:%d\n", __func__, __LINE__, errno);
-        return -1;
+        close(lsock);
```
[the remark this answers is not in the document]
Yashu37: Thanks for the review. Updated the listen() failure path to use the existing close_listener_socket() cleanup helper, which properly closes the listener socket and removes the Unix socket path. The recv() failure path also closes m_dsock. Existing logging and control flow are preserved.

src/agent/em_cmd_agent.cpp
```diff
@@ -66,7 +66,8 @@ int em_cmd_agent_t::execute(em_long_string_t result)
 
     if ((ret = listen(lsock, 20)) == -1) {
         printf("%s:%d: listen error on socket, err:%d\n", __func__, __LINE__, errno);
-        return -1;
+        close_listener_socket(lsock, get_svc());
```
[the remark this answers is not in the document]
Yashu37: Thanks for the review. Updated the listen() failure path to pass em_service_type_agent to close_listener_socket() instead of get_svc(). This ensures the listener socket and Unix socket path are properly cleaned up in LOCAL_CLI builds.

src/agent/em_cmd_agent.cpp
```diff
@@ -55,7 +66,8 @@ int em_cmd_agent_t::execute(em_long_string_t result)
 
     if ((ret = listen(lsock, 20)) == -1) {
         printf("%s:%d: listen error on socket, err:%d\n", __func__, __LINE__, errno);
-        return -1;
+        close_listener_socket(lsock, em_service_type_agent);
```
Nikita-Hakai, line 69: indentation mismatch 
can you also change printf to em_printfout

Yashu37: Fixed the indentation mismatch and changed printf() to em_printfout() as suggested.
Thank you.

src/agent/em_cmd_agent.cpp
```diff
@@ -54,8 +65,9 @@ int em_cmd_agent_t::execute(em_long_string_t result)
     }
 
     if ((ret = listen(lsock, 20)) == -1) {
-        printf("%s:%d: listen error on socket, err:%d\n", __func__, __LINE__, errno);
-        return -1;
+        em_printfout("%s:%d: listen error on socket, err:%d\n", __func__, __LINE__, errno);
+        close_listener_socket(lsock, em_service_type_agent);
+	return -1;
```
Nikita-Hakai:
```suggestion
        return -1;
```

src/agent/em_cmd_agent.cpp
```diff
@@ -74,7 +86,8 @@ int em_cmd_agent_t::execute(em_long_string_t result)
 
         if ((ret = recv(m_dsock, tmp, sizeof(em_event_t) + EM_MAX_EVENT_DATA_LEN, 0)) <= 0) {
             printf("%s:%d: listen error on socket, err:%d\n", __func__, __LINE__, errno);
-            break;
+            close(m_dsock);
+	    break;
```
Nikita-Hakai:
```suggestion
            break;
```

src/agent/em_cmd_agent.cpp
Yashu37: Fixed the indentation mismatch.
Thank you.

src/agent/em_cmd_agent.cpp
Yashu37: Fixed the indentation mismatch.
Thank you.

Nikita-Hakai (approved):

PR #802, opened 2026-09-12
RDKCOM-5649: RDKBDEV-3507 Fix socket FD leak in em_cmd_exec_t::get_listener_socket()

Yashu37:
**Issue:**
In em_cmd_exec_t::get_listener_socket(), a listener socket is successfully created using socket(). If the subsequent bind() operation fails, the function returns without closing the socket descriptor. This leaves the socket file descriptor open and can cause a resource/FD leak when the failure occurs repeatedly.

**Steps to Reproduce:**
    1.Invoke em_cmd_exec_t::get_listener_socket().
    2.Allow the listener socket to be created successfully.
    3.Trigger a failure in the bind() operation.
    4.Observe that the function returns -1 without closing the created socket descriptor.
    5.Repeat the failure and observe the accumulation of open socket descriptors.

**Expected Behavior:**
When bind() fails, the listener socket descriptor should be closed before returning an error.

**Fix:**
Add close(lsock) in the bind() failure path before return -1.

**Acceptance Criteria:**
1.lsock is closed when bind() fails.
2.No socket file descriptor leak occurs on the bind() failure path.
3.Existing successful socket creation and binding behavior remains unchanged.
4.Existing error handling and return values are preserved.


src/cmd/em_cmd_exec.cpp
```diff
@@ -201,7 +201,8 @@ int em_cmd_exec_t::get_listener_socket(em_service_type_t svc)
 
     if ((ret = bind(lsock, reinterpret_cast<const struct sockaddr *> (&addr), sizeof(struct sockaddr_un))) == -1) {
         printf("%s:%d: bind error on socket: %d, err:%d\n", __func__, __LINE__, lsock, errno);
-        return -1;
+        close(lsock);
+	return -1;
```
Nikita-Hakai:
```suggestion
        return -1;
```

Yashu37: Fixed the indentation mismatch.
Thank you.

Nikita-Hakai (approved):

PR #803, opened 2026-09-12
RDKCOM-5654: RDKBDEV-3508 Fix socket file descriptor leak in em_cmd_exec_t::send_cmd()

Yashu37:
**Issue:**
In em_cmd_exec_t::send_cmd(), the destination socket dsock is successfully created using socket(). If the subsequent connect() operation fails, the function returns -1 without closing dsock. This leaves the socket file descriptor open and can cause a resource leak when the failure occurs repeatedly.

**Steps to Reproduce:**
    1.Invoke em_cmd_exec_t::send_cmd().
    2.Allow the destination socket to be created successfully.
    3.Trigger a failure in the connect() operation.
    4.Observe that the function returns -1 without closing dsock.
    5.Repeat the failure and observe accumulation of open socket descriptors.

**Expected Behavior:**
When connect() fails, the socket file descriptor should be closed before returning an error.

**Fix:**
Add close(dsock) in the connect() failure path of em_cmd_exec_t::send_cmd() before return -1.

**Acceptance Criteria:**
1.dsock is closed when connect() fails.
2.No socket file descriptor leak occurs on the connect() failure path.
3.Existing successful connection behavior remains unchanged.
4.Existing error handling and return values are preserved.


src/cmd/em_cmd_exec.cpp
```diff
@@ -328,7 +328,8 @@ int em_cmd_exec_t::send_cmd(em_service_type_t to_svc, unsigned char *in, unsigne
 
     if ((ret = connect(dsock, reinterpret_cast<const struct sockaddr *> (&addr), sizeof(struct sockaddr_un))) != 0) {
         snprintf(out, out_len, "%s:%d: connect error on socket, err:%d\n", __func__, __LINE__, errno);
-        return -1;
+        close(dsock);
+	return -1;
```
Nikita-Hakai, line 332: indentation mismatch

Yashu37: Fixed the indentation mismatch.
Thank you.

Nikita-Hakai (approved):

